PR #6, opened 2025-09-20
Optimise requirements.txt and update pipecat imports

a6kme:
1. Fix imports for updated pipecat
2. Optimise requirement.txt


chewwbaka (approved):
LGTM. Note -
- updated pipecat commit hash in requirements.txt
- added webrtc in pipecat-ai dependencies as suggested

PR #8, opened 2025-09-22
feat: Trickle ice candidates for faster WebRTC connection

a6kme:
Instead of gathering all ice candidates before generating SDP response, let the ICE trickle to and from UI.


chewwbaka:
wrong branch.

PR #35, opened 2025-10-24
Feat/telephony-integrations

chewwbaka:
- generalise telephony integrations
- add vonage telephony integration


api/alembic/versions/a57d25b75117_add_provider_info_to_cost_info.py
```diff
@@ -0,0 +1,122 @@
...
+            CASE 
+                WHEN cost_info IS NULL OR cost_info::text = '{}' 
+                THEN '{}'::jsonb 
+                ELSE cost_info::jsonb 
+            END,
+            '{provider}', 
+            '"twilio"'::jsonb,
+            true
+        )::json
+        WHERE mode = 'twilio'
```
a6kme: I am not sure what this migration is supposed to do? What will we set as value of provider cost in the `cost_info`?

api/routes/organization.py
```diff
@@ -1,50 +1,102 @@
...
-async def get_telephony_configuration(user: UserModel = Depends(get_user)):
-    """Get telephony configuration for the user's organization with masked sensitive fields."""
+async def get_telephony_configuration(
+    user: UserModel = Depends(get_user),
+    provider: Optional[str] = None  # Query param to filter by provider
+):
+    """Get telephony configuration for the user's organization with masked sensitive fields.
+    
+    Args:
+        provider: Optional provider filter ('twilio' or 'vonage').
```
a6kme: Is there a provision to store multiple providers at once? When a user has an existing configuration as Twilio, and the user changes the configuration to Vonage from the UI, will we store both configuration or only the updated configuration?

If we only store the latest configuration, the parameter `provider` is redundant.

api/routes/organization.py
```diff
@@ -53,33 +105,73 @@ async def save_telephony_configuration(
...
-        # Check if auth_token is unchanged (masked value matches)
-        stored_auth_token = existing_config.value.get("auth_token", "")
-        if stored_auth_token and is_mask_of(request.auth_token, stored_auth_token):
-            config_value["auth_token"] = stored_auth_token  # Keep original
-
+        existing_provider = existing_config.value.get("provider")
+        
+        # Only preserve masked values if it's the same provider
+        if existing_provider == request.provider:
+            if request.provider == "twilio":
```
a6kme: Nit: These parts can be DRYd up

api/routes/telephony.py
```diff
@@ -0,0 +1,507 @@
...
+                user_id
+            )
+            
+    except Exception as e:
+        logger.error(f"Error in WebSocket connection: {e}")
+        await websocket.close(1011, "Internal server error")
+
+
+@router.post("/status-callback/{workflow_run_id}")
+async def handle_status_callback(
```
a6kme: I would also recommend to name the endpoints to reflect the provider. Examples

`/status-callback/{provider}/{workflow_run_id}`

and this can handle provider specific status updates. The provider specific status update logic can remain in the provider classes to colocate the provider specific implementation.

PR #135, opened 2026-01-28
fix: BACKEND_API_ENDPOINT resolution from env and cloudflared tunnel

chewwbaka:
- in the absence of this env variable on a local oss setup, cloudflared tunnel url is used for the url formation and reverse proxy handles the https scheme
- the env variable is needed when its a custom domain

### DIFF
--- a/api/utils/common.py
+++ b/api/utils/common.py
@@ -1,0 +1,181 @@
+"""
+Common utilities.
+Shared functions used across the application.
+"""
+
+import re
+
+from loguru import logger
+
+from api.constants import BACKEND_API_ENDPOINT
+from api.utils.tunnel import TunnelURLProvider
+
+
+def get_scheme(url: str) -> str | None:
+    """
+    Extract scheme from a given URL if present.
+    Returns None if not found
+    """
+    idx = url.find("://")
+    if idx == -1:
+        return None
+    return url[:idx]
+
+
+def _validate_url(url: str) -> None:
+    """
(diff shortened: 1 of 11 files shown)


api/utils/common.py
```diff
@@ -0,0 +1,106 @@
...
+    env_endpoint = os.getenv("BACKEND_API_ENDPOINT")
+    
+    if env_endpoint:
+        logger.debug(f"Processing BACKEND_API_ENDPOINT from environment: {env_endpoint}")
+        
+        # Handle localhost special case - use tunnel URL instead
+        if "localhost" in env_endpoint:
+            logger.debug(f"BACKEND_API_ENDPOINT is localhost ({env_endpoint}), using tunnel URL instead")
+            # Second priority: Query cloudflared tunnel URL
+            return await TunnelURLProvider.get_tunnel_urls()
```
a6kme: This will fail if we have localhost as environment variable and cloudflare is not setup.

api/utils/common.py
```diff
@@ -0,0 +1,106 @@
...
+        
+        try:
+            # Parse the URL to validate and handle protocol
+            parsed = urlparse(env_endpoint)
+            
+            if parsed.scheme:
+                # Case 1: URL has protocol (http or https) - use as is
+                if parsed.scheme in ['http', 'https']:
+                    http_url = urlunparse((
```
a6kme: when we are reconstructing the URL like this, this can be easily replaced with `http_url = env_endpoint`. Both are essentially equivalent.

PR #155, opened 2026-02-13
feat: telephony call transfer

chewwbaka:
- call transfer only for telephony calls
- supports twilio telephony only


docker-compose.yaml
```diff
@@ -74,87 +74,87 @@ services:
...
 
-  api:
-    image: ${REGISTRY:-dograhai}/dograh-api:latest
-    volumes:
-      - shared-tmp:/tmp
-    environment:
-      # Core application config
-      ENVIRONMENT: "local"
-      LOG_LEVEL: "INFO"
+  # api:
```
a6kme: @chewwbaka - We might have to undo these changes.

PR #171, opened 2026-02-23
feat: tansfer calls with aasterisk

chewwbaka:


api/services/workflow/pipecat_engine_custom_tools.py
```diff
@@ -516,8 +538,29 @@ async def transfer_call_handler(
...
+                        self._cleanup_transfer_context_delayed(transfer_id)
+                    )
+
         return transfer_call_handler
 
+    async def _cleanup_transfer_context_delayed(self, transfer_id: str):
+        """Background task to clean up transfer context after pipeline processing delay."""
+        try:
+            # Wait for pipeline to process EndFrame(reason="transfer_call") in serializers
+            await asyncio.sleep(1.0)  # 1 second delay for async pipeline processing
```
a6kme: This seems like code smell. Can we think of something else?

api/services/telephony/ari_manager.py
```diff
@@ -633,6 +710,124 @@ async def _delete_bridge(self, bridge_id: str):
...
+        """Handle transfer destination channel answered - publish success event."""
+        try:
+            logger.info(
+                f"[ARI Transfer org={self.organization_id}] Destination {destination_channel_id} "
+                f"answered for transfer {transfer_id}"
+            )
+
+            # Store channel mapping for potential future events and get transfer context
+            transfer_manager = await self._get_transfer_manager()
+            await transfer_manager.store_transfer_channel_mapping(
```
a6kme: If we are storing destination channel id with transfer id, why do we need to call `_find_transfer_context_for_call`?

api/services/telephony/providers/ari_call_strategies.py
```diff
@@ -0,0 +1,253 @@
...
+    async def _find_transfer_context_for_call(self, caller_channel_id: str):
+        """Find the active transfer context for this caller channel."""
+        try:
+            import redis.asyncio as aioredis
+
+            from api.constants import REDIS_URL
+            from api.services.telephony.transfer_event_protocol import TransferContext
+
+            # Search Redis for transfer contexts where original_call_sid matches this caller
+            redis = aioredis.from_url(REDIS_URL, decode_responses=True)
```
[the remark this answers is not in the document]
chewwbaka: done

api/services/telephony/ari_manager.py
chewwbaka: store_transfer_channel_mapping has the destination channel id. we have these mappings in redis-
1. destination channel id <> transfer_id. this comes in handy in stasis events for the first leg of call transfer (outbound channel to destination number)
2. transfer id <> context
3.  _find_transfer_context_for_call is used when we only have the caller call id (in asterisk serliazer) & involves iterating on all the contexts to match on caller call id. I can optimize this by maintaing one more mapping of caller channel id <> transfer_id.
Let me check this and update accordingly.

PR #187, opened 2026-03-10
feat: add CAMB AI TTS integration

neilruaro-camb:
## Summary
- Add CAMB AI as a TTS provider with `mars-flash`, `mars-pro`, and `mars-instruct` models
- Register `CambTTSConfiguration` in the service registry with voice ID and language fields
- Wire up CAMB branch in `create_tts_service()` factory using pipecat's `CambTTSService`
- Add API key validator stub and update Dockerfile with `camb` pipecat extra


api/services/configuration/registry.py
```diff
@@ -391,6 +377,21 @@ class SarvamTTSConfiguration(BaseTTSConfiguration):
...
+
+@register_tts
+class CambTTSConfiguration(BaseTTSConfiguration):
+    provider: Literal[ServiceProviders.CAMB] = ServiceProviders.CAMB
+    model: str = Field(
+        default="mars-flash", json_schema_extra={"examples": CAMB_TTS_MODELS}
+    )
+    voice: str = Field(default="147320", description="Camb.ai voice ID")
+    language: str = Field(default="en-us", description="BCP-47 language code")
+    api_key: str
```
chewwbaka: Remove api_key: str override. The base class BaseServiceConfiguration already defines api_key: str | list[str] and handles list-to-string conversion. Overriding with api_key: str causes a       
  Pydantic validation error when the UI submits the key as a single-element list (which is how all TTS provider keys are submitted).

PR #189, opened 2026-03-13
Feat multiple did

Atulptl:


chewwbaka:
Hi, we have a different approach for this functionality in mind, together with more architectural changes around telephony configuration as a whole. Closing this PR, please discuss over Slack if theres an urgency at your end that we could find a workaround for.

Issue #254, opened 2026-04-29
[Bug]: Few Telnyx calls have static and bot is unable to respond to user replies


chewwbaka:
### Describe the bug

- Telnyx calls with UK number are failing with a lot of static and bots inability to reply to users audio messages

### Steps to Reproduce

1. Configure Telnyx telephony, add a UK phone number
2. Make a telephony call with Telnyx as the default configuration
3. Note the static in dograh call recordings
4. Check the recordings at telnyx by enabling the call recording in outbound voice profile

### Expected behavior

1. There shouldn't be any static in the call

PR #256, opened 2026-04-29
fix: honor telnyxs per-call codec in bidirectional stream

chewwbaka:
Telnyx's bidirectional stream uses different codecs per direction:
  - Dograh → Telnyx: what we declare via `stream_bidirectional_codec`
  - Telnyx → Dograh: whatever the PSTN leg negotiated (PCMA for UK, Europe, India termination; PCMU for US), announced as `media_format.encoding` on the WebSocket start message.

We hardcoded both directions on the serializer to PCMU, so any call whose PSTN leg used PCMA arrived as A-law bytes that we decoded through a μ-law table → static for the entire call.

`handle_websocket` now extracts `media_format.encoding` from the start message and threads it through `transport_kwargs`. The serializer uses it for the Telnyx → Dograh direction; the Dograh → Telnyx direction stays pinned to PCMU to match the unchanged `stream_bidirectional_codec` in the dial/answer payloads.

Note: pipecat's `TelnyxFrameSerializer` names its params from the call's POV, not Dograh's — `inbound_encoding` is what we send into the call, `outbound_encoding` is what we receive. Easy to mix up.

### DIFF
--- a/api/services/telephony/providers/telnyx/provider.py
+++ b/api/services/telephony/providers/telnyx/provider.py
@@ -82,6 +82,9 @@ class TelnyxProvider(TelephonyProvider):
             f"{backend_endpoint}/api/v1/telephony/telnyx/events/{workflow_run_id}"
         )
 
+        # stream_bidirectional_codec controls only the Dograh → Telnyx direction.
+        # The Telnyx → Dograh direction follows the PSTN leg and is announced via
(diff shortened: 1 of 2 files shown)


chewwbaka:
fixes #254

a6kme (approved):

PR #271, opened 2026-05-09
Verify Telnyx webhook signatures

a692570:
## Summary

This PR implements Telnyx webhook signature verification for the shared inbound dispatcher and the Telnyx status event route. It replaces the Telnyx accept-all stub with Ed25519 verification against the stored `webhook_public_key`.

## Changes

- Builds the signed payload as `{timestamp}|{raw_request_body}`, matching Telnyx webhook signing docs.
- Reads `telnyx-signature-ed25519` and `telnyx-timestamp` from request headers.
- Uses the configured Telnyx `webhook_public_key`, not a request-provided public key.
- Fails closed when the signature, timestamp, public key, or verification material is missing or invalid.
- Verifies `/telnyx/events/{workflow_run_id}` before processing status updates, including streaming events.
- Adds `PyNaCl` for Ed25519 verification.
- Adds tests for valid signatures, tampered bodies, missing headers, missing public key config, case-insensitive headers, and status-route verification before processing.

## Out of scope

- No `/telnyx/inbound/run` route. The shared `/inbound/run` dispatcher already detects Telnyx, matches `connection_id`, verifies the webhook, and starts the inbound stream.
- No `WEBHOOK_ENDPOINT` change. Telnyx status callbacks are sent through the `webhook_url` supplied on Call Control dial or answer commands, not through the generic markup-response endpoint.
- No `get_call_cost` implementation. I did not find a documented single-call `/v2/cdr/calls` lookup by `call_leg_id`.
- No API key examples or credential-like fixtures.

## References

- [Telnyx webhook signing](https://developers.telnyx.com/development/api-fundamentals/webhooks/receiving-webhooks)
- [Telnyx Call Control dial webhooks and streaming fields](https://developers.telnyx.com/api-reference/call-commands/dial)
- [Telnyx answer command `webhook_url` and streaming fields](https://developers.telnyx.com/api-reference/call-commands/answer-call)
- [Telnyx Call Control Application `webhook_event_url`](https://developers.telnyx.com/api-reference/call-control-applications/update-a-call-control-application)
- [Telnyx usage reports](https://developers.telnyx.com/docs/reporting/usage-reports)

## Security

- No real API keys, tokens, passwords, private keys, or internal hostnames are included.
- Tests generate Ed25519 keys at runtime and use placeholder webhook payloads only.
- The public key is treated as saved configuration and is never accepted from the incoming request.

## Testing

```bash
../venv/bin/python -m pytest tests/telephony/telnyx/test_provider.py -q
```

Result: 7 passed, 1 warning.

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -18,3 +18,4 @@ bcrypt==5.0.0
 email-validator==2.3.0
 posthog==7.11.1
 fastmcp==3.2.4
+PyNaCl==1.6.2
--- a/api/services/telephony/providers/telnyx/provider.py
+++ b/api/services/telephony/providers/telnyx/provider.py
@@ -4,11 +4,15 @@ Uses the Telnyx Call Control API v2 for outbound calling with
 inline WebSocket media streaming.
 """
 
+import base64
+import binascii
 import json
 import random
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 import aiohttp
+import nacl.exceptions
+import nacl.signing
 from fastapi import HTTPException, WebSocketDisconnect
 from loguru import logger
 
@@ -32,6 +36,13 @@ def normalize_event_type(event_type: str) -> str:
     dotted form so all downstream matching can use a single canonical shape.
     """
     return (event_type or "").replace("_", ".")
+
+
+def _get_header(headers: Dict[str, str], name: str) -> str:
+    for key, value in headers.items():
+        if key.lower() == name:
+            return value
+    return ""
 
 
 class TelnyxProvider(TelephonyProvider):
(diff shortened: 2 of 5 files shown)


chewwbaka:
Hi @a692570, thanks for the PR - the change looks good. I pushed a few hardening fixes on top; happy to discuss any of them.
 
- The timestamp was part of the signed payload but never validated for freshness- added a 5-minute replay window matching the constant Telnyx uses across all of its official SDKs. 
- Invalid signatures returned implicit HTTP 200, which hid failures from both Telnyx's retry logic and our HTTP-level metrics -- changed to 401, plus 400 for non-UTF-8 bodies and 404 for unknown workflows. 
- Combined log messages conflated failure modes (e.g. "missing signature or timestamp")- split them one-line-per-mode and bumped operator-side public-key misconfig to error level. 
- Negative tests cover the new error paths.

a692570:
Thanks for the hardening fixes make sense to me.

PR #274, opened 2026-05-12
feat(telephony/telnyx): add call transfer via conference bridge

chewwbaka:
Conference-based transfer for Telnyx with a two-step flow:
- transfer_call dials the destination with a per-transfer webhook URL.
- On call.answered, the webhook seeds a conference with the destination's live call_control_id and publishes DESTINATION_ANSWERED.
- TelnyxConferenceStrategy joins the caller into the conference on pipeline teardown (EndTaskReason.TRANSFER_CALL).
- On post-answer destination hangup, the webhook hangs up the caller — Telnyx's create_conference doesn't accept end_conference_on_exit on the seed leg, so we tear down the bridge ourselves.

TransferContext gains optional workflow_run_id (for webhook→provider resolution in multi-config orgs) and conference_id (set on answer, rd by the strategy).

Also fixes the transfer tool's provider lookup to go through get_telephony_provider_for_run instead of the deprecated org-default shim, which was returning the wrong provider in multi-config orgs.

### DIFF
--- a/api/services/telephony/providers/telnyx/provider.py
+++ b/api/services/telephony/providers/telnyx/provider.py
@@ -620,8 +620,127 @@ class TelnyxProvider(TelephonyProvider):
         timeout: int = 30,
         **kwargs: Any,
     ) -> Dict[str, Any]:
-        """Telnyx call transfer is not yet implemented."""
-        raise NotImplementedError("Call transfer not yet supported for Telnyx")
+        """Dial the destination as a plain call; conference is seeded later.
+
+        Webhook (``call.answered``) seeds the conference with this leg;
+        ``TelnyxConferenceStrategy`` joins the caller on pipeline teardown.
+        https://developers.telnyx.com/api-reference/call-commands/dial
+        """
+        if not self.validate_config():
+            raise ValueError("Telnyx provider not properly configured")
+
+        from_number = random.choice(self.from_numbers)
+        logger.info(f"Selected phone number {from_number} for Telnyx transfer call")
+
+        backend_endpoint, _ = await get_backend_endpoints()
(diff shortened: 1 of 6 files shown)


chewwbaka:
For the offline review comment about the need of introducing conference_id for telnyx when we could have reused conference_name
1. conference_name is created in pipecat_engine_custom_tools.py, so it's present for every provider. It's the natural human label for observability/tracing across all transfer flows.                                                                                                
2. Each provider uses a different operational handle for the conference:                                                                                                                                                                                                                                          
  - Twilio: the name we set in TwiML is the handle - all conference actions reference it by name.                                                                                                                                                                                                                   
  - Telnyx: POST /v2/conferences returns a UUID id; every subsequent action endpoint (/v2/conferences/{id}/actions/...) requires that id, not the name.                                                                                                                                                             
3. Overloading conference_name to also carry the Telnyx id would mean the same field holds different values at different points in the journey (a name pre-create, a UUID post-create). We'd also lose the human name, which is useful for tracing and is a consistent convention across providers.                 
4. The Telnyx flow uses conference_id for idempotency on duplicate call.answered webhooks: if conference_id is None, create the conference; otherwise skip. Overloading the name field breaks this clean boolean check - we'd be left to inspect the string shape (UUID vs. our transfer-{id} pattern), which is heuristic.                                                                                                                                                                                                                                                                                                          
5. Possible cleaner evolution: rename conference_id -> conference_handle (provider-agnostic "operational handle"). Both providers populate both conference_name and conference_handle. For Twilio they're equal; for Telnyx, name is the human label and handle is the UUID. Out of scope for this PR but a good follow-up.

Issue #283, opened 2026-05-13
[Bug]: Stasis bridge creation failing intermittently for asterisk ari telephony calls


chewwbaka:
### Describe the bug

- There is complete silence in asterisk calls both outbound and inbound sometimes. 
- The bridge creation between channel and external media is failing as sometimes when the bridge is being created the state of external media is still down(websocket connection isnt established yet). its a race condition between bridge being created and websocket connection opening(for the external media channel) 

### Steps to Reproduce

1. Make both outbound and inbound calls with asterisk ari telephony

### Expected behavior

1. call should work everytime

a6kme:
Fixed by #284

PR #284, opened 2026-05-13
fix(ari): pre-register ext channel id and defer bridge to its StasisS…

chewwbaka:
…tart

Two race conditions in the inbound ARI flow could leave a call silent:

1. Bridging both channels immediately after creating the ext media leg raced against the ext channel entering the Stasis application; slow chan_websocket handshakes produced "Channel not in Stasis application" 422 errors on addChannel.

2. Asterisk could fire StasisStart for the ext channel before the externalMedia POST response returned, so _is_ext_channel returned False and the event was dropped as an unknown outbound call.

Fixes:
- Generate the ext channel id as dograh-ext-<uuid> client-side and pass it to Asterisk via the channelId query param. Mark the ext channel, set its channel->run mapping, register the pending bridge entry, and persist gathered_context.ext_channel_id all before the POST.
- Defer the bridge to a new _complete_bridge_after_ext_ready handler triggered by the ext channel's own StasisStart. Both channels are guaranteed in Stasis by then, so addChannel cannot 422.
- On POST failure or channelId mismatch, roll back the pending entry and ERROR loudly.

### DIFF
--- a/api/services/telephony/ari_manager.py
+++ b/api/services/telephony/ari_manager.py
@@ -36,7 +36,9 @@ from api.services.telephony.transfer_event_protocol import (
 # Redis key pattern and TTL for channel-to-run mapping
 _CHANNEL_KEY_PREFIX = "ari:channel:"
 _EXT_CHANNEL_KEY_PREFIX = "ari:ext_channel:"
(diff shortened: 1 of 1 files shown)


chewwbaka:
Fixes #283

PR #309, opened 2026-05-18
feat: add MiniMax provider support (Chat + TTS)

octo-patch:
## Summary

Add MiniMax as a new provider for both LLM (Chat) and TTS services.

**Chat (LLM)**:
- Uses MiniMax's OpenAI-compatible API (`https://api.minimax.io/v1`)
- Models: `MiniMax-M2.7`, `MiniMax-M2.7-highspeed`
- Authentication via `MINIMAX_API_KEY`
- Temperature default set to 1.0 (MiniMax doesn't accept 0)

**TTS**:
- Leverages Pipecat's built-in `MiniMaxHttpTTSService`
- Models: `speech-2.8-hd` (default), `speech-2.8-turbo`
- 6 built-in voice options
- Requires `group_id` configuration

## Changes

- `api/services/configuration/registry.py`: Add `MINIMAX` provider enum, LLM config class, TTS config class
- `api/services/pipecat/service_factory.py`: Add factory branches for LLM and TTS
- `api/tests/test_minimax_service_factory.py`: Unit tests for both providers

## API Documentation

- Chat (OpenAI Compatible): https://platform.minimax.io/docs/api-reference/text-openai-api
- TTS: https://platform.minimax.io/docs/api-reference/speech-t2a-http
- Pricing: https://platform.minimax.io/docs/guides/pricing-paygo

### DIFF
--- a/api/services/configuration/check_validity.py
+++ b/api/services/configuration/check_validity.py
@@ -53,6 +53,7 @@ class UserConfigurationValidator:
             ServiceProviders.ASSEMBLYAI.value: self._check_assemblyai_api_key,
             ServiceProviders.GLADIA.value: self._check_gladia_api_key,
             ServiceProviders.RIME.value: self._check_rime_api_key,
+            ServiceProviders.MINIMAX.value: self._check_minimax_api_key,
         }
 
     async def validate(
@@ -146,6 +147,19 @@ class UserConfigurationValidator:
             except ValueError as e:
                 return [{"model": service_name, "message": str(e)}]
             return []
+
+        # MiniMax TTS requires a group_id alongside the API key.
+        # LLM configs don't expose group_id, so only check when the field exists.
+        if provider == ServiceProviders.MINIMAX.value and hasattr(
+            service_config, "group_id"
+        ):
+            if not getattr(service_config, "group_id", None):
+                return [
+                    {
+                        "model": service_name,
+                        "message": "group_id is required for MiniMax TTS",
+                    }
+                ]
 
         api_key = service_config.api_key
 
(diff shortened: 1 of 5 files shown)


chewwbaka:
Hi @octo-patch, thanks for the PR! 
Pushed a few follow-up commits with the fixes below to land this quickly. Please flag anything that doesn't match your intent.                                                                                                                                                   
                                                                                                                                                                                                                                                                                         
Correctness fixes:                                                                                                                                                                                                                                            
- Validator plumbing for MiniMax
- temperature=1.0 was hardcoded in create_llm_service_from_provider
- TTS aiohttp session was leaking — aiohttp.ClientSession() was created in create_tts_service but never closed. Wrapped MiniMaxHttpTTSService in MiniMaxOwnedSessionTTSService (new module api/services/pipecat/minimax_tts.py) that closes the session in cleanup(). 
- TTS base_url requires /t2a_v2 — pipecat appends only ?GroupId=... to whatever you pass. Added base_url to MiniMaxTTSConfiguration (default https://api.minimax.io/v1/t2a_v2) so users in mainland China / US-West can override                                                         
- Reasoning model <think> leak — MiniMax-M2.7 emits <think>...</think> blocks inline in delta.content (no separate reasoning_content field — confirmed by capturing the raw stream). Without filtering, the reasoning text gets aggregated into sentences and sent to TTS. Added         
MiniMaxLLMService in pipecat that strips the leading think block from the stream and re-emits it as LLMThought*Frame events
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
Intentionally not included (so you know we considered them):                                                                                                                                                                                                            
- Exposing volume, pitch, text_normalization, language_boost, emotion on the TTS config - emotion had no audible effect on the voices we tested (English_Graceful_Lady, English_expressive_narrator) and skipped the rest for now; happy to take this up as a follow-up.
                                                                                                                                                                                                                                                                                              
Merging once CI is green unless you flag something. Anything we should follow up on, let's track in a separate PR.

PR #312, opened 2026-05-18
fix(gemini-live): seed conversation history on node-transition reconnect

VincentEmmanuel:
## Summary

- Multi-node workflows on Gemini Live often produce "Sorry, I didn't catch that" / "I haven't heard anything from you yet" on the LLM's first turn after a node transition, even when the user clearly spoke.
- Root cause: the reconnect triggered by a `system_instruction` change passes a stale `session_resumption_handle` (issued before the user's most recent message), so the new server-side session restores state from before the user spoke.
- Fix: discard the handle on node-transition reconnects in both `_handle_changed_settings` (sync) and `_set_bot_is_responding` (deferred), then seed history from `self._context` in `_handle_session_ready` via `_create_initial_response(for_reconnect=True)`.

See the commit message for the full root-cause writeup and the rationale for the seed-shape choice (Gemini 3.x `turn_complete=False` + reliance on the function-call-result frame to trigger the first turn).

## Repro before this fix

Any two-node workflow on Gemini Live 3.x (e.g. `startCall` → `agentNode`). Caller hears the greeting, says something substantive, and the first turn in the second node is "Sorry, didn't catch that" — confirmed by inspecting `rtf-bot-text` events on the run record (the user's transcription is final and present, but the LLM in the new session doesn't see it).

Reproduced in production on `gemini-3.1-flash-live-preview` across ~6 consecutive runs with different workflow shapes.

## Test plan

- [x] `pytest api/tests/test_gemini_live_reconnect_history_seed.py` — 5 new cases (handle-discard in sync + deferred paths, seed-on-reconnect-without-handle, first-connect path unchanged, defensive no-op)
- [x] `pytest api/tests/test_gemini_live_reconnect_tool_results.py` — existing tests still pass
- [x] End-to-end on a deployed Dograh instance against `gemini-3.1-flash-live-preview`: pre-fix → "I haven't heard anything from you yet"; post-fix → LLM engages directly with the user's pre-transition message.
- [ ] Reviewer to spot-check that real-resumption-handle reconnects (network blip recovery) are unaffected — those still fall through both branches in `_handle_session_ready`.

## What this doesn't touch

- `_create_initial_response` itself is unchanged. The seed shape (turn_complete=False for Gemini 3.x reconnect) was already the documented intended behavior; we just weren't reaching it.
- Tool-result delivery on reconnect (covered by `test_gemini_live_reconnect_tool_results.py`) is unaffected.
- The fix is scoped to `DograhGeminiLiveLLMService` — pipecat's upstream `GeminiLiveLLMService` is untouched.

### DIFF
--- a/api/services/pipecat/realtime/gemini_live.py
+++ b/api/services/pipecat/realtime/gemini_live.py
@@ -82,6 +82,15 @@ class DograhGeminiLiveLLMService(GeminiLiveLLMService):
             # cut the bot off mid-utterance.
             self._reconnect_pending = True
         else:
+            # Discard session_resumption_handle before a node-transition
+            # reconnect. The handle is issued periodically by Gemini, not
+            # after every transcription — so resuming to it restores state
+            # from BEFORE the user's most recent message. The new session's
+            # LLM then doesn't see what the user just said and replies with
+            # "I haven't heard anything from you yet". Clearing the handle
+            # forces a fresh server-side session that _handle_session_ready
+            # will fully seed from self._context.
+            self._session_resumption_handle = None
             await self._reconnect()
         return {"system_instruction"}
 
@@ -110,6 +119,10 @@ class DograhGeminiLiveLLMService(GeminiLiveLLMService):
             await self._run_pending_function_calls()
             if self._reconnect_pending:
                 self._reconnect_pending = False
+                # Same rationale as in _handle_changed_settings: discard the
+                # handle so the deferred node-transition reconnect uses a
+                # fresh session that we can fully seed.
+                self._session_resumption_handle = None
                 await self._reconnect()
 
     async def _run_pending_function_calls(self):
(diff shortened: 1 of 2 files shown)


api/services/pipecat/realtime/gemini_live.py
```diff
@@ -82,6 +82,15 @@ async def _handle_changed_settings(self, changed: dict[str, Any]) -> set[str]:
...
         else:
+            # Discard session_resumption_handle before a node-transition
+            # reconnect. The handle is issued periodically by Gemini, not
+            # after every transcription — so resuming to it restores state
+            # from BEFORE the user's most recent message. The new session's
+            # LLM then doesn't see what the user just said and replies with
+            # "I haven't heard anything from you yet". Clearing the handle
+            # forces a fresh server-side session that _handle_session_ready
+            # will fully seed from self._context.
+            self._session_resumption_handle = None
```
a6kme, line 93: Hey @VincentEmmanuel - Thank you for your PR. 

The flow currently is 
1. LLM Calls for a node switch
2. The executor sets the system instructions with new node
3. We reconnect with existing session resumption handler so that old context is in the model on reconnect
4. We send the tool response from `_process_completed_function_calls` when we see the `ContextFrame` pushed by Assistant Aggregator when it sees `FunctionCallResultFrame`. 

> The handle is issued periodically by Gemini, not after every transcription — so resuming to it restores state from BEFORE the user's most recent message.

Could you give me some evidence of this? In my mind, the state restored is that - Gemini has sent a tool call and its awaiting the response of that tool call. 

I understand that with this change, you wish to bootstrap a new session with the messages in our context. Could you help me with some concrete reason for why that would be better than our current approach? Happy to hear any adjacent thoughts you might have on this. Thanks!

a6kme:
Closing due to inactivity.

PR #313, opened 2026-05-18
feat: add OpenAI Realtime async tool support

alexandreolives:
### Summary

Adds OpenAI Realtime async tool support while keeping the existing sync provider flow unchanged.

### Why

OpenAI Realtime can keep speaking while tool calls are still running. This improves latency and makes the conversation flow feel more natural.

### What changed

- Added a Dograh wrapper for OpenAI Realtime.
- Enabled async tool cancellation for OpenAI Realtime sessions.
- Added support for the `function_call_output` payload shape expected by OpenAI Realtime.
- Kept synchronous providers unchanged, including Google realtime.

### Workaround

This PR includes a small Dograh-side compatibility workaround for the current Pipecat OpenAI Realtime tool flow.

Until upstream Pipecat fully matches the async serialization expected by OpenAI Realtime, Dograh:
- tracks async tool calls while they are running
- emits a running marker for in-flight async tools
- sends the final result back in the OpenAI Realtime payload format

This workaround is scoped to OpenAI Realtime only and does not affect sync providers.

### Tests

- `api/tests/test_realtime_provider_defaults.py`
- `api/tests/test_realtime_message_append.py`
- `api/tests/test_openai_realtime_initial_context.py`
- `api/tests/test_run_pipeline_realtime_turn_config.py`

### Notes

- Google realtime keeps the existing non-async behavior.
- The async path is only enabled for OpenAI Realtime.
- The workaround can be simplified or removed later if upstream Pipecat fully covers the same behavior.


a6kme:
Hello @alexandreolives - Thank you for your contribution. 

We actually gate the function call executions behind `BotStoppedSpeakingFrame`. The idea is that if the llm wants to execute a tool while speaking something, and gets interrupted by the user, I feel a safer thing to do is to not execute the tool at all. 

Supporting async tools is something I need to give more thought into.

a6kme:
Closing due to inactivity.

PR #314, opened 2026-05-18
fix(telephony/ari): force softmix bridge to fix chan_websocket one-way audio

tecnomanu:
## Summary

Fixes one-way audio (agent → user works, user → agent does not) on outbound calls using the **Asterisk ARI provider with `chan_websocket`** external media. Verified end-to-end with Asterisk 20.19 + PJSIP trunk: after this change, user audio finally reaches the Dograh VAD/pipeline.

Closes the symptoms described in #307.

## Root cause

`_create_bridge_and_add_channels` creates the bridge with only `type=mixing` and then adds exactly 2 channels (PJSIP caller + chan_websocket external media), both ulaw, no framehook, no recording. Under those conditions Asterisk's bridging core selects `simple_bridge` (P2P native) as the bridge technology.

`simple_bridge` + `chan_websocket` has a well-known asymmetry: frames flow from the WebSocket leg to the PJSIP leg (the agent's TTS reaches the user) but **not** in the reverse direction. The user can hear the agent; the agent never hears the user.

Verified at the wire level: `pjsip show channelstats` showed `Receive Count` growing at ~50pps while the user spoke (audio reaching Asterisk's PJSIP leg) but Dograh's VAD only logged `no audio received while speaking`.

Confirmed it is specific to `simple_bridge` + `chan_websocket`: the same Asterisk container with `chan_audiosocket` instead worked bidirectionally without changes.

## Fix

Add `dtmf_events,proxy_media,video_sfu` to the bridge `type`. The `video_sfu` capability is the decisive one — `simple_bridge` cannot satisfy video routing, so the bridging core has no choice but to pick `softmix`, which handles both audio directions correctly.

Verified with `asterisk -rx "bridge show all"`:

**Before:**
```
Bridge-ID  Name                Chans Type    Technology      Duration
2bb73ecc.. bridge-1779138763.3     2 stasis  simple_bridge   00:00:53
```

**After:**
```
Bridge-ID  Name                Chans Type    Technology      Duration
5d2b631b.. bridge-1779138944.5     2 stasis  softmix         00:00:01
```

User audio now reaches the agent pipeline, conversation is fully bidirectional.

## Trade-offs

`softmix` is heavier than `simple_bridge` (it actually mixes frames in software instead of relaying them P2P), so there's a small CPU cost per call. For the 2-channel case it's marginal and worth it given that simple_bridge is fundamentally broken with chan_websocket. Native bridge was never an intentional optimization for this code path — it was incidental, the result of not specifying any other capability.

## Things I considered but did not do

- **Forcing transcoding via codec asymmetry** — fragile, depends on trunk configuration.
- **Adding a silent 3rd channel** — more invasive, needs originating a Local channel.
- **Installing a framehook (JITTERBUFFER, MixMonitor)** — side effects on the call leg.

The single-line change to the bridge `type` parameter is the minimal, side-effect-free fix.

## Test plan

- [x] Tested against Asterisk 20.19 + PJSIP ulaw trunk + chan_websocket external media on a real outbound call.
- [x] Verified bridge technology is `softmix` after the change (was `simple_bridge`).
- [x] Verified bidirectional audio (user speech reaches VAD, agent responds).
- [x] No regression on agent → user direction (TTS still reaches the user).

Happy to add automated coverage if there's an existing test pattern for ARI bridge creation — I didn't find one in the repo.

### DIFF
--- a/api/services/telephony/ari_manager.py
+++ b/api/services/telephony/ari_manager.py
@@ -490,7 +490,11 @@ class ARIConnection:
             "POST",
             "/bridges",
             params={
-                "type": "mixing,dtmf_events,proxy_media",
+                # Force softmix selection by including video_sfu — simple_bridge
+                # cannot satisfy video routing capabilities, so the bridging core
+                # picks softmix. Required because simple_bridge + chan_websocket
+                # has a known asymmetry where PJSIP->WS frames don't flow.
+                "type": "mixing,dtmf_events,proxy_media,video_sfu",
                 "name": f"bridge-{channel_ids[0]}",
             },
         )
--- a/api/services/telephony/ari_manager.py
+++ b/api/services/telephony/ari_manager.py
@@ -489,7 +489,10 @@ class ARIConnection:
         bridge_result = await self._ari_request(
             "POST",
             "/bridges",
-            params={"type": "mixing", "name": f"bridge-{channel_ids[0]}"},
+            params={
+                "type": "mixing,dtmf_events,proxy_media",
+                "name": f"bridge-{channel_ids[0]}",
+            },
         )
         bridge_id = bridge_result.get("id", "")
         if not bridge_id:


chewwbaka:
Hi @tecnomanu , thanks for the detailed write-up. 

Before merging, I want to flag one thing worth retesting.                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                          
You tried rtp_keepalive=5 on the trunk endpoint and saw no effect. There's a known [Asterisk bug](https://github.com/asterisk/asterisk/issues/359) where rtp_keepalive can silently stop firing after the first packet, depending on its value relative to rtcpinterval in rtp.conf (default 5000 ms). The exact rule isn't clearly documented, but rtp_keepalive=5 is right on the edge — could explain why your test looked like it did nothing.                                                                                                                                                                                                                                                                                         
   
Could you retest with rtp_keepalive=2 on the trunk endpoint, leaving rtcpinterval at its default? A value of 2 sits safely below the bug's threshold under any reading of the rule.                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                        
The reason I ask - [a community thread](https://community.asterisk.org/t/audio-transmission-issues-with-simple-bridge-in-ari/105332) describes a very similar symptom you hit - PJSIP + external media -> simple_bridge → no inbound audio - and resolved it with rtp_keepalive alone. (I believe you were trying to link the same thread in your issue description?)
Your VPS deployment with a public-internet SIP trunk is the kind of setup where stateful firewalls on the trunk leg can drop idle RTP - what rtp_keepalive addresses. If keepalive does resolve it, the root cause is likely a firewall/NAT closing the inbound RTP path, and softmix is masking that by continuously generating outbound frames.                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                          
Two possible outcomes:                                                                                                                                                                                                                                                                                                                

1. rtp_keepalive=2 resolves it → we'd prefer to document the keepalive requirement in our ARI integration docs rather than pay softmix CPU on every call.                                                       
2. Still fails → softmix really is the right fix and we'll merge this, plus add the keepalive docs as belt-and-suspenders.                                                                                                                                                                                                                                                                                                                

Either way, your investigation is what made this tractable - thanks for the thorough PR and the issue!

chewwbaka:
Hi @tecnomanu - going to close this for now due to inactivity. Happy to reopen if you can share updates described below.                                               
                                            
Summary of what we verified locally:                                                                                                                     
- The PR's mechanism works on Asterisk 20.19.0: type=mixing,dtmf_events,proxy_media,video_sfu does force softmix selection over simple_bridge.         
- We could not reproduce the user -> agent audio failure in our test environment.                                                                           
                                                                              
Open question we couldn't close: whether rtp_keepalive=2 on the trunk endpoint resolves the symptom in your setup. Your earlier =5 may have been         
undermined by Asterisk issue #359 (rtp_keepalive can silently stop firing depending on its value relative to rtcpinterval).                              
                                        
If you have time to retest with rtp_keepalive=2 or share additional captures from your environment, please reopen.                                       
                                                                                                                                                         
Thanks for the thorough investigation - interesting issue to work on even though we couldn't close it together.

tecnomanu:
Hi @chewwbaka, sorry for the delay. I ran the `rtp_keepalive=2` test you suggested. Result: bug reproduces identically, softmix is the right fix.

**Setup** (clean reproduction):

- Reverted the `ari_manager.py` patch — bridge type back to `"mixing"` only
- Removed a `set_var=JITTERBUFFER(adaptive)=default` I had on the trunk endpoint (was another softmix-forcing workaround via framehook — same effect as the bridge type change)
- Set `rtp_keepalive=2` on the trunk endpoint
- Confirmed `rtcpinterval` is at default (not set in `rtp.conf`)
- Verified `pjsip show endpoint trunk-astervoip` reports `rtp_keepalive: 2`
- Bridge confirmed as `simple_bridge` during the call (Asterisk verbose log: `Channel PJSIP/trunk-astervoip-... joined 'simple_bridge' stasis-bridge`)

**Result**: bug reproduces, user audio never reaches the pipeline.

Evidence from a 111s call:

- `VADController: no audio received while speaking` fires 4 times in 14 seconds while I'm actively talking
- `PipelineTask: heartbeat frame not received for more than 10.0 seconds` fires ~15 times after the 10s mark — input frames stop entirely
- `pjsip show channelstats` during the call:

\`\`\`
BridgeId  ChannelId             UpTime    Codec  Recv:Count Lost Pct  Xmit:Count
1240d2ec  trunk-astervoip-...   00:01:54  ulaw   521        117  22%  4475
\`\`\`

  Even those 521 received packets don't reach the WS channel — VAD/heartbeat logs show the bridge isn't forwarding them.
- Call completes normally at SIP/RTP transport level (no timeouts, no disconnects). The break is strictly at the bridge layer.

So `rtp_keepalive=2` doesn't address it: packets do arrive at the PJSIP leg, `simple_bridge` simply doesn't pass them to chan_websocket. Confirms the asymmetry is the real root cause, not NAT/firewall.

For reference, my deployment setup using the patched fork is at https://github.com/tecnomanu/dokploy-dograh/tree/test/ws-fix — that branch builds the images from this PR's fork branch and runs them via Dokploy. With the patch applied I have stable bidirectional audio in production; without it (this test) user→agent breaks consistently.

Could you reopen the PR?

tecnomanu:
@chewwbaka please remember check this PR and reopen. Thanks!

PR #315, opened 2026-05-19
docs: clarify MCP available tools and reduce tool-name confusion

Bortlesboat:
## Summary
- add an explicit Available MCP tools section listing all currently registered server tools
- clarify that assistants should stick to registered tool names
- normalize a few UI/docs punctuation artifacts for clearer rendering

## Why
There is active confusion around missing MCP tool names. A canonical list in MCP docs helps users and agents avoid calling nonexistent tools.

## Validation
- reviewed pi/mcp_server/server.py tool registration and mirrored exact names in docs

PR #316, opened 2026-05-19
feat(mcp): add search_docs tool over docs corpus (closes #295)

### LINKED ISSUE #295: [Bug]: docs-search tool does not exist in MCP
### Describe the bug

The dograh MCP documentation mentions that you can search the documentation using MCP, but the search docs tool does not exist in MCP. 

### Steps to Reproduce

Image

### Expected behavior

We should have a docs search MCP tool

### Screenshots

_No response_

### Device details

_No response_

### Additional context

_No response_

SuperMarioYL:
## Summary

The docs at https://docs.dograh.com promise *"Search the Dograh docs for how to configure a TURN server"* as an MCP example prompt, but no `search_docs` tool exists — agents can list workspace resources but cannot search the documentation. Issue #295.

This adds a dependency-free, in-process keyword search over the `docs/` tree that's already shipped into the API image (`COPY ./docs ./docs` in `api/Dockerfile`).

## Changes

| File | Change |
|---|---|
| `api/mcp_server/tools/docs_search.py` | **new** — `async search_docs(query, limit=10)` |
| `api/mcp_server/server.py` | register `search_docs` in the tool list (+2 lines) |
| `api/tests/test_mcp_docs_search.py` | **new** — 18 unit tests, mocks `authenticate_mcp_request` |

## Behaviour

- Weighted keyword scoring: path > title > body. A page whose title IS the query term outranks one that merely mentions it (`title_weight=5` vs `body_cap=4 × body_weight=1`).
- Hard 25-result cap regardless of caller-supplied `limit`; default 10 keeps the MCP payload small.
- Snippet extraction (~240 chars) centered on the first term hit, whitespace collapsed for inline MCP rendering.
- Page-title resolution: Mintlify YAML frontmatter → first ATX heading → filename fallback.
- URL building strips `.mdx`/`.md` and collapses `index` so `getting-started/index.mdx` → `https://docs.dograh.com/getting-started`.
- Docs root discovery: `DOGRAH_DOCS_PATH` env override → `/app/docs` (Docker) → walk-up from `__file__` (source/dev).
- Authentication: required, for consistency with other MCP tools and to route through the existing rate-limit middleware. Docs themselves are not org-scoped.
- Graceful degradation: returns `[]` if the corpus isn't on disk rather than 500-ing.
- Input validation: empty/punct-only queries and `limit < 1` raise `ValueError` with actionable messages.

## Why no vector index

The docs corpus is ~100 files / ~140k LoC. A per-call scan runs comfortably under 50 ms, so adding an embedding backend, vector store, or background indexer would be a heavier dependency than the problem warrants — especially for self-hosted/offline deployments. The tool is called interactively from MCP, not from a hot path.

## Tests

`api/tests/test_mcp_docs_search.py` covers:

- **Pure helpers** — tokenizer rejects 1-char and punctuation-only terms; frontmatter stripping; title extraction (frontmatter → heading → filename); URL building (`.mdx` strip + `/index` collapse); scoring weights (title-outweighs-body invariant); env-override resolution.
- **End-to-end** — TURN-server query ranks the dedicated page first against a synthetic corpus; non-doc files (`docs.json`) are excluded; no-match queries return `[]`; `limit=1` honored; `limit=999` clamped to 25; empty-corpus returns `[]`; invalid queries (empty / punct-only / `limit=0`) raise `ValueError`.

Each test is mocked at `authenticate_mcp_request` to match the pattern in `test_mcp_save_workflow.py` — no live DB/Redis required.

Lint: `ruff check` + `ruff format --check` pass on all three files.

## Test plan

- [x] `ruff check api/mcp_server/tools/docs_search.py api/mcp_server/server.py api/tests/test_mcp_docs_search.py` — clean
- [x] `ruff format --check ...` — already-formatted
- [x] End-to-end logic verified manually against a synthetic 4-file corpus (TURN-server query returns the right top hit; non-doc files excluded; limit clamping works; degenerate-input errors raise)
- [ ] `pytest api/tests/test_mcp_docs_search.py` — should run green in CI (couldn't run locally without postgres/redis from `api/.env.test`)

## Related

- Issue #295 — the bug this closes
- PR #315 — complementary docs-side change ("remove the promise from docs"); this PR takes the other path ("fulfill the promise")

If you'd prefer the docs-side fix as the canonical resolution, feel free to close this — I went the tool-add route because the example prompts in `docs/integrations/mcp.mdx` ("Search the Dograh docs for how to configure a TURN server") read like a real intended feature rather than aspirational wording.


a6kme:
Hello @SuperMarioYL - Thank you for your contribution. 

I have done some changes in the approach, and also avoided the case when there might be drift between the documentation and the code. 

Example: if someone is pinned on v1.20 on their self hosted, they should only refer to the docs corresponding to that version in the running instance rather than docs.dograh.com, which might have docs from version 1.25 and may drift from the capabilities of v1.20. 

I am pushing everything in the branch and merging this PR. Hope thats okay.

PR #317, opened 2026-05-19
fix(stt): align Speechmatics language registry with official transcription codes

palinko91:
I wasn't able to use my language for transcription and I've realised Speechmatics list was outdated. I've updated it according to the newest documentation I've found:

https://docs.speechmatics.com/speech-to-text/languages#transcription-languages


a6kme:
Thanks for your contribution @palinko91 🚀

Issue #297, opened 2026-05-15
[Bug]: Remote server install not working on Proxmox host


rubenclark74:
### Describe the bug

Install succeeds using the remote server install script. All containers start successfully, and I can bring up the web interface and log in. However, whenever I test an agent web call, it never connects and ends the call after a few seconds.

### Steps to Reproduce

- Create new Docker LXC on Proxmox.
- Install Dograh using provided script:
     `curl -o setup_remote.sh https://raw.githubusercontent.com/dograh-hq/dograh/main/scripts/setup_remote.sh && chmod +x setup_remote.sh && ./setup_remote.sh`
- Configure LAN IP when asked (192.168.5.195)
- Browse to https://192.168.5.195 and log in
- Create new agent and test

### Expected behavior

The test call should connect, but never does. There is no audio. I see the following for a few moments:

<img width="550" height="341" alt="Image" src="https://github.com/user-attachments/assets/2b4674b6-6cd9-4e7a-850d-aa5bbc861f55" />

Then afterwards I see it processing the call, even though a connection was never established.

okphotouk:
same issue here the browser/client is appearing as your public WAN IP in the logs , while Dograh is internal-only. So WebRTC is trying to connect via TURN/NAT and then dropping.

a6kme:
@okphotouk / @rubenclark74 - I think I might know whats happening. We are filtering out local IPs in the SDP to cut noise for the Turn Server. https://github.com/dograh-hq/dograh/blob/2381a803ade54f6c8d1db572e0f6c3301dd74c20/api/routes/webrtc_signaling.py#L95

Could you try setting the environment variable `FORCE_TURN_RELAY` and let me know if it works for you guys? If it works, I will figure out something for deployment on local ip addresses. If you guys have some ideas around, would be happy to hear them out.

a6kme:
Meanwhile, I have pushed the change in the script to set it automatically if it detects a private ip during setup. 

https://github.com/dograh-hq/dograh/blob/fc04f31639e0d326525d6840ca117babe2b25ea8/scripts/setup_remote.sh#L52

rubenclark74:
Ok, I have set the variable. Now it hangs on 'establishing connection' and does not appear to time out, but still does not connect.

a6kme:
[...]

PR #321, opened 2026-05-19
Web rtc lan fix

deepashreeKedia:
## Fix: WebRTC Voice Call Failure (LAN Setup)

  **Symptom:** Voice calls fail on LAN setup.

  ---
  
  ### Fix 1 — Relay candidate silently dropped by `filter_outbound_sdp`

  coturn runs on the same LAN server, so its relay candidate IP is private. The private IP filter in
  `filter_outbound_sdp` was dropping it, leaving the browser with zero server candidates.

  **Fix:** Exempt relay candidates from the private IP check — a relay IP being private is expected in LAN deployments.

  ---

  ### Fix 2 — NAT hairpin failure in `_handle_ice_candidate`

  With only the browser's `srflx` candidate (the router's public IP) allowed through, the server routed STUN checks via 
  TURN to that public IP. The router doesn't hairpin the packet back to the browser on the LAN, so ICE never completes. 
  The browser's host candidate would have given the server a direct LAN path via TURN, but it was being dropped.

  **Fix:** Bypass the private IP filter when `FORCE_TURN_RELAY` is on — cloud production never enables this flag, so the 
  filter remains intact there.


a6kme:
Thanks @deepashreeKedia for the PR. Fixed #297

PR #361, opened 2026-05-26
chore: upgrade Next.js in evals/visualizer from 16.1.4 to 16.2.6

### LINKED ISSUE #338: Best-practice: outdated Next.js (middleware-bypass advisories), duplicate lockfiles, fail-open OSS_JWT_SECRET default
Hi dograh team,

While testing [AI PatchLab](https://github.com/elfrost/ai-patchlab) (an open-source local-first SAST/SCA scanner) on a range of AI/agent projects, I scanned dograh at \`21951eca18cd\`. The scan is small and tells one clean headline story plus a few minor items — filing as a single courtesy issue.

Full curated write-up with the FP analysis: https://elfrost.github.io/ai-patchlab/scans/dograh-hq-dograh.html

## 1. Outdated Next.js across \`ui/\` and \`evals/visualizer/\` — one upgrade clears ~45 findings

\`next\` is pinned to a vulnerable version in three lockfiles (\`ui/package-lock.json\`, \`evals/visualizer/package-lock.json\`, \`evals/visualizer/pnpm-lock.yaml\`). Trivy surfaces the full advisory set. By priority:

- **Middleware / proxy bypass** — multiple advisories (Pages Router i18n, App Router segment-prefetch, dynamic route parameter injection, plus an "incomplete fix follow-up"). This is the class worth fixing first: Next.js middleware is commonly where auth/authorization checks live, so a bypass can mean an unauthenticated request reaching a protected route. For a voice-agent platform whose UI fronts campaign/contact data, that's directly in the threat path.
- **SSRF via WebSocket upgrades.**
- **DoS variants** (connection exhaustion via Cache Components / Server Components, Image Optimization API).
- **XSS variants** (\`beforeInteractive\` scripts, CSP-nonce App Router) and **cache poisoning** in RSC responses.

None of these is a dograh code bug — it's the standard Next.js advisory backlog that accumulates when the framework pin drifts. Fix: upgrade \`next\` to the current patched release in each front-end, re-lock, and rebuild the \`dograh-ui\` container image so the upgrade actually reaches deployments.

## 2. \`evals/visualizer/\` ships two lockfiles (\`package-lock.json\` **and** \`pnpm-lock.yaml\`)

Both are present and both are out of date (the Next.js advisories fire against both). Whichever lockfile a given developer/CI honours, the other goes stale silently. Pick one package manager for the sub-project and delete the other lockfile.

## 3. \`docker-compose.yaml:180\` — \`OSS_JWT_SECRET\` defaults to \`ChangeMeInProduction\`

\`\`\`yaml
OSS_JWT_SECRET: "\${OSS_JWT_SECRET:-ChangeMeInProduction}"
\`\`\`

The naming makes the intent clear, which is good — but \`\${VAR:-default}\` fails **open**: a deployment that never sets \`OSS_JWT_SECRET\` runs with the JWT signing key literally being the string \`ChangeMeInProduction\`, a value committed in a public repo. Anyone can then forge JWTs against that deployment.

The one-character fix makes it fail **closed**:

\`\`\`yaml
OSS_JWT_SECRET: "\${OSS_JWT_SECRET:?OSS_JWT_SECRET must be set}"
\`\`\`

\`\${VAR:?message}\` aborts \`docker compose up\` with the message if the variable is unset — the deployment fails loudly instead of silently running with a known-public signing key.

## 4. \`api/Dockerfile\` — minor hardening

- No \`USER\` directive — the API container runs as root. Add a non-root user before the entrypoint.
- \`apt-get install\` omits \`--no-install-recommends\` — adding it shrinks the image and the package attack surface.

---

Happy to open PRs for any of these. Thanks for dograh — the scan was clean overall (the only false positives were API-doc token placeholders in \`docs/**/*.mdx\` and the PostHog \`phc_\` public project key in \`docker-compose.yaml\`, which is write-only and designed to ship). The Next.js upgrade is the one item really worth prioritizing.

nuthalapativarun:
## Summary

`evals/visualizer/package.json` pinned `next: 16.1.4`, which has known security advisories including middleware/proxy bypass, SSRF, DoS, and reflected XSS issues. This upgrades to `16.2.6` (latest at time of writing) and updates the `pnpm-lock.yaml` accordingly. The evals visualizer is an internal dev tool with no external exposure, but keeping dependencies current reduces supply-chain risk.

## Changes

- `evals/visualizer/package.json`: `next` 16.1.4 → 16.2.6, `eslint-config-next` 16.1.4 → 16.2.6
- `evals/visualizer/pnpm-lock.yaml`: Regenerated lockfile with updated resolutions
- `evals/visualizer/pnpm-workspace.yaml`: Removed `allowBuilds` placeholder stubs added by `pnpm up`

## Testing

`pnpm up next@latest eslint-config-next@latest` ran successfully and resolved to 16.2.6. The `ignoredBuiltDependencies` for `sharp` and `unrs-resolver` are preserved to avoid native build failures in CI.

Closes #338


a6kme:
@nuthalapativarun - Thanks for your contribution. 🚀

PR #362, opened 2026-05-26
fix: stamp API key into model override at save time to survive global provider change

### LINKED ISSUE #336: [Bug]: Model Override fails if there is no global configuration
### Describe the bug

If we have a model override for provider A and we dont have a global config for that model, the calls will fail. We rely on Global Config to get API Key. A potential solution is to copy the API Keys in model configurations. 

### Steps to Reproduce

- Set model A in global TTS with voice X
- Override that model in agent with model A and voice Y
- Set model B in global TTS
- Try to make a call

### Expected behavior

The override should not break

### Screenshots

_No response_

### Device details

_No response_

### Additional context

_No response_

nuthalapativarun:
## Summary

When a workflow overrides the TTS/LLM/STT provider to the same provider as the current global config (e.g. ElevenLabs voice Y over global ElevenLabs voice X), the override dict stores only model/voice/language fields — not the API key. The override relies on the global config to supply the key at resolve time.

If the user later changes the global TTS to a different provider (e.g. Deepgram), `resolve_effective_config` detects the provider mismatch and tries to build a new ElevenLabs config from the override dict alone — which fails because there is no API key in the override. Calls break silently.

**Fix**: Add `enrich_overrides_with_api_keys()` to `resolve.py`. At workflow-save time, this function detects when an override's provider matches the current global section, and stamps the global API key (and other secret fields like `credentials`, `aws_access_key`) into the override dict. The override becomes self-contained and survives future global provider changes.

## Changes

- `api/services/configuration/resolve.py`: Add `enrich_overrides_with_api_keys()` that copies secret fields from the global config into the model_overrides dict for matching providers
- `api/routes/workflow.py`: Call `enrich_overrides_with_api_keys()` in the workflow update handler before saving, so the enriched overrides are persisted to DB
- `api/tests/test_resolve_effective_config.py`: Add `TestEnrichOverridesWithApiKeys` including a regression test that verifies the full bug scenario end-to-end

## Testing

Added five tests including a regression test that reproduces the exact steps from the issue:
1. Global TTS = ElevenLabs (with API key)
2. Override TTS = ElevenLabs (different voice, no API key)
3. `enrich_overrides_with_api_keys` stamps the ElevenLabs key into the override
4. Global TTS changes to Deepgram
5. `resolve_effective_config` with the enriched override correctly returns ElevenLabs config with the preserved API key

Closes #336

### DIFF
--- a/api/routes/workflow.py
+++ b/api/routes/workflow.py
@@ -21,9 +21,11 @@ from api.sdk_expose import sdk_expose
 from api.services.auth.depends import get_user
 from api.services.configuration.check_validity import UserConfigurationValidator
 from api.services.configuration.masking import (
+    mask_workflow_configurations,
     mask_workflow_definition,
     merge_workflow_api_keys,
 )
+from api.services.configuration.merge import merge_workflow_configuration_secrets
 from api.services.configuration.resolve import (
     enrich_overrides_with_api_keys,
     resolve_effective_config,
@@ -412,7 +414,9 @@ async def create_workflow(
         "current_definition_id": workflow.current_definition_id,
         "template_context_variables": workflow.template_context_variables,
         "call_disposition_codes": workflow.call_disposition_codes,
-        "workflow_configurations": workflow.workflow_configurations,
+        "workflow_configurations": mask_workflow_configurations(
+            workflow.workflow_configurations
+        ),
     }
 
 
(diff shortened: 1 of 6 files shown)


a6kme:
Thanks for your contribution @nuthalapativarun. I have added some more tests and added the masking and merging logic so that the API keys dont leak to different consumers of the workflow definition.

PR #377, opened 2026-05-28
feat: allow overriding base URL of OpenAI STT and TTS

developer603:
## What

Mirrors the LLM treatment from #368 for the OpenAI **STT** and OpenAI **TTS** providers. Users running OpenAI-compatible self-hosted services (vLLM, llama.cpp, custom proxies) can now point Dograh at them via the OpenAI provider with `base_url`, instead of being forced onto the Speaches provider as a workaround.

## Why

#368 added `base_url` for OpenAI **LLM**. Self-hosted operators who run their own STT/TTS on the same OpenAI-compatible API surface (it's the same convention pipecat already supports — both `OpenAISTTService` and `OpenAITTSService` constructors accept `base_url`) have to keep using the Speaches provider as a sidecar. With this change, the OpenAI provider becomes a clean, uniform path for all three service types.

## Changes

- **`registry.py`** — adds `base_url` field (default `https://api.openai.com/v1`) to `OpenAISTTConfiguration` and `OpenAITTSService`, identical in shape to the `OpenAILLMService.base_url` from #368.
- **`service_factory.py`** — in the OPENAI branches of `create_stt_service` and `create_tts_service`, lift `base_url` off `user_config`, run it through `_validate_runtime_service_url`, and forward it as a kwarg. Same pattern as the LLM branch from #368.
- **`test_user_configured_service_url_security.py`** — adds 4 runtime tests covering private-IP rejection and localhost rejection in SaaS mode for both STT and TTS.

`check_validity.py` requires no edit — the existing per-service URL validation loop already iterates `("base_url", "endpoint")` via `getattr`, and `_check_openai_api_key` already routes OPENAI providers through the base_url-aware code path (also from #368).

## Compatibility

- No schema migration. Existing configurations without `base_url` keep their default and continue to talk to `api.openai.com`.
- OSS-mode behaviour unchanged (URL validator is a no-op when `DEPLOYMENT_MODE=oss`).
- SaaS-mode tightening matches LLM: private/localhost/CGNAT IPs rejected with the same `validate_user_configured_service_url` helper.

## Tests

```
$ pytest api/tests/test_user_configured_service_url_security.py
23 passed in 4.80s   (19 existing + 4 new)
```

## Related

Pattern source: #368 (LLM base_url support). This PR completes the STT + TTS legs.

### DIFF
--- a/api/services/configuration/registry.py
+++ b/api/services/configuration/registry.py
@@ -803,6 +803,10 @@ class OpenAITTSService(BaseTTSConfiguration):
         default="alloy",
         description="OpenAI TTS voice name.",
     )
+    base_url: str = Field(
+        default="https://api.openai.com/v1",
+        description="Override only if using an OpenAI-compatible API (e.g. local TTS, proxy).",
+    )
 
 
 DOGRAH_TTS_MODELS = ["default"]
@@ -1060,6 +1064,10 @@ class OpenAISTTConfiguration(BaseSTTConfiguration):
         default="gpt-4o-transcribe",
         description="OpenAI transcription model.",
         json_schema_extra={"examples": OPENAI_STT_MODELS},
+    )
+    base_url: str = Field(
+        default="https://api.openai.com/v1",
+        description="Override only if using an OpenAI-compatible API (e.g. local STT, proxy).",
     )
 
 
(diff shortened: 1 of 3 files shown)


a6kme:
@developer603 - Thank you for your contribution.

PR #397, opened 2026-06-01
fix: make email lookup case-insensitive in get_user_by_email

developer603:
# fix: make email lookup case-insensitive in `get_user_by_email`

## Problem
Email addresses are case-insensitive in practice, but `UserClient.get_user_by_email`
matched with an exact predicate:

```python
select(UserModel).where(UserModel.email == email)
```

A user who signed up as `User@example.com` is not found when they later log in as
`user@example.com` (and vice-versa). Depending only on how the client capitalized
the address, the same person can fail to authenticate — or be treated as a brand-new
account. This surfaces in the wild when users sign in from different
devices/clients/SSO flows that normalize capitalization differently.

## Fix
Compare on the lowercased email so lookups are robust to capitalization:

```python
select(UserModel).where(func.lower(UserModel.email) == func.lower(email))
```

## Why this approach
- **Minimal & backwards-compatible:** works with existing mixed-case rows
  immediately — no data migration required.
- **No write-path change:** storage is untouched, so nothing else in the codebase
  needs to be aware of the change.

## Possible follow-ups (out of scope, noted for maintainers)
- Add a functional index `CREATE INDEX ... ON users (lower(email))` so the
  case-insensitive lookup stays index-backed at scale.
- Optionally normalize email to lowercase on `create_user_with_email` /
  `update_user_email` to keep a single canonical form (would want a one-time
  backfill + a uniqueness guard against pre-existing case-variant duplicates).

## Testing
- `python -m py_compile` clean.
- Verified logically: a row stored as `User@Example.com` is returned for inputs
  `user@example.com`, `USER@EXAMPLE.COM`, and `User@Example.com`.


a6kme:
Thanks @developer603 for your contribution.

PR #398, opened 2026-06-01
fix: detect masked short secrets in contains_masked_key

developer603:
# fix: detect masked short secrets in `contains_masked_key`

## Problem
`contains_masked_key()` guards against persisting a still-masked secret by testing
for the `MASK_MARKER` (`"***"`) substring:

```python
return any(MASK_MARKER in k for k in keys)
```

But `mask_key()` only produces 3+ consecutive asterisks for keys longer than
`VISIBLE_CHARS + 2`. For short secrets it produces fewer:

```python
mask_key("EMPTY")  # -> "*MPTY"   (a single asterisk)
mask_key("X")      # -> "*"
```

These masked values slip past the guard, so a dashboard **"save without editing"**
round-trips the masked display string back to the server and overwrites the real
stored value with e.g. `"*MPTY"` — silently corrupting the secret.

This is most visible for self-hosted / OpenAI-compatible providers that use a short
no-validate sentinel api_key such as `"EMPTY"` (common for local vLLM, Ollama, etc.).

## Fix
Match the full shape `mask_key()` actually produces — a run of `MASK_CHAR` followed
by at most `VISIBLE_CHARS` revealed trailing characters — in addition to the legacy
marker:

```python
_MASK_SHAPE_RE = re.compile(
    rf"^{re.escape(MASK_CHAR)}+[^{re.escape(MASK_CHAR)}]{{0,{VISIBLE_CHARS}}}$"
)
...
return any(MASK_MARKER in k or bool(_MASK_SHAPE_RE.match(k)) for k in keys)
```

## Testing
Verified against `mask_key()` output and against real unmasked values:

| input (real) | `mask_key()` | old detector | new detector |
|---|---|---|---|
| `EMPTY` | `*MPTY` | ❌ False | ✅ True |
| `X` | `*` | ❌ False | ✅ True |
| `mykey` | `*ykey` | ❌ False | ✅ True |
| `sk-1234567890abcdef` | `***************cdef` | ✅ True | ✅ True |
| `abcd` | `****` | ✅ True | ✅ True |

Real **unmasked** values are not flagged — no false positives:
`contains_masked_key("EMPTY") / ("sk-live-abcd") / ("mykey") / ("X")` all return `False`
(no leading-asterisk run, so the anchored regex doesn't match).

`python -m py_compile` clean.

### DIFF
--- a/api/services/configuration/masking.py
+++ b/api/services/configuration/masking.py
@@ -10,6 +10,7 @@ The rules are simple:
 """
 
 import copy
+import re
 from typing import Any, Dict, Optional
 
 from api.schemas.user_configuration import UserConfiguration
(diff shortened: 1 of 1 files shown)


chewwbaka:
Hi @developer603 nice ctach and thanks for fixing the short masked-secret case. I noticed one edge case that still seems to fail with the new detection logic:

```python
masked = mask_key("vsk*v")  # "*sk*v"
contains_masked_key(masked)  # False
```

Since mask_key() can reveal * in the trailing visible characters, this is still a valid masked value and should be detected. Could you adjust the detection logic and add a regression test for this case?

chewwbaka:
HI @developer603 I’m going to close this PR since we’re landing the change separately in another branch/PR.

The main missing piece was the edge case from the review comment:

```python
mask_key("vsk*v")  # "*sk*v"

mask_key() can reveal * in the visible suffix, so the detector needs to match the actual mask shape rather than excluding * from the trailing visible characters.

Follow-up PR with the adjusted implementation and regression coverage: https://github.com/dograh-hq/dograh/pull/482

PR #407, opened 2026-06-03
fix(telephony): handle Cloudonix CDR webhooks missing session/disposition

Mubashirrrr:
## Bug

`POST /api/v1/telephony/cloudonix/cdr` is a public, unauthenticated webhook (no `Depends`, no signature check) that parses arbitrary external JSON. It unconditionally dereferences:

```python
call_id = cdr_data.get("session").get("token")
```

If a CDR payload omits `session` (or sends `"session": null`), `cdr_data.get("session")` is `None`, so `None.get("token")` raises `AttributeError` → unhandled **HTTP 500**. The graceful guard right below it is unreachable because the crash happens first:

```python
call_id = cdr_data.get("session").get("token")   # crashes here
if not call_id:                                   # never reached when session missing
    return {"status": "error", "message": "Missing call_id field"}
```

`StatusCallbackRequest.from_cloudonix_cdr()` (called later in the same handler) has the same `session` fragility plus a second one: `data.get("disposition", "")` returns `None` when the key is present-but-null, and `None.upper()` then crashes.

## Reproduce

```python
def route_logic(cdr_data):
    if not cdr_data.get("domain"):
        return {"status": "error", "message": "Missing domain field"}
    call_id = cdr_data.get("session").get("token")   # line 106
    if not call_id:
        return {"status": "error", "message": "Missing call_id field"}
    ...

route_logic({"domain": "acme.cloudonix.io", "disposition": "ANSWER"})
# AttributeError: 'NoneType' object has no attribute 'get'
```

```python
StatusCallbackRequest.from_cloudonix_cdr({"session": {"token": "t"}, "disposition": None})
# AttributeError: 'NoneType' object has no attribute 'upper'
```

## Fix

Navigate both fields defensively so missing/null values fall through to the intended graceful error path instead of crashing:

- `routes.py`: `(cdr_data.get("session") or {}).get("token")`
- `status_processor.py`: `disposition = data.get("disposition") or ""` and `call_id=(data.get("session") or {}).get("token") or ""`

Surgical — every changed line maps to the crash. Well-formed payloads map exactly as before.

## Regression test

`api/tests/telephony/test_cloudonix_cdr.py` covers: CDR route with missing `session`, route with null `session`, `from_cloudonix_cdr` with missing/null `session`+`disposition`, and the well-formed mapping path. Each route case asserts the graceful `{"status": "error", "message": "Missing call_id field"}` instead of a 500. Confirmed failing (AttributeError) before the fix and passing after.


chewwbaka:
Hi @Mubashirrrr, Thanks for the fix. The regression coverage for missing/null `session` and `disposition` is clear.
Handled for when session is a truthy non-object value, e.g. `"session": "abc"`, which would still raise on `.get("token")`.

PR #408, opened 2026-06-03
fix(qa): tolerate non-dict JSON from QA LLM instead of crashing

Mubashirrrr:
## Bug

`parse_llm_json` (`api/services/gen_ai/json_parser.py`) is **explicitly designed to return a list** when the model emits a top-level JSON array — it has a dedicated unit test for `"[1, 2, 3]"`, and `_extract_json_array` is one of its parse strategies.

The QA analyzers in `api/services/workflow/qa/analysis.py` (both `_run_qa_analysis` per-node and the `_run_whole_call_qa_analysis` fallback) call `.get()` directly on the result:

```python
try:
    parsed = parse_llm_json(raw_response)
    node_result["tags"] = parsed.get("tags", [])      # AttributeError if parsed is a list
    node_result["summary"] = parsed.get("summary", "")
    ...
except (json.JSONDecodeError, ValueError):
    ...
```

When `parsed` is a list, `parsed.get(...)` raises `AttributeError`, which is **not** caught by the surrounding `except (json.JSONDecodeError, ValueError)`. So a single stray array response from the QA model crashes the entire QA analysis run instead of degrading to empty results.

Notably, the live variable-extraction path already guards this exact case (`api/services/workflow/pipecat_engine.py`: `if not isinstance(extracted_data, dict): ... skip`), so this is an inconsistency — the same defensive check is present in one consumer of LLM-parsed JSON but missing in the QA consumers.

## Reproduce

```python
from api.services.gen_ai.json_parser import parse_llm_json
import json

raw_response = '["tag1", "tag2"]'   # model returned a top-level array
node_result = {}
try:
    parsed = parse_llm_json(raw_response)
    node_result["tags"] = parsed.get("tags", [])     # boom
except (json.JSONDecodeError, ValueError):
    node_result["tags"] = []
# AttributeError: 'list' object has no attribute 'get'
```

## Fix

Mirror the existing engine guard in both QA call sites: coerce a non-dict parse result to `{}` so the `.get()` lookups produce empty defaults instead of crashing.

```python
parsed = parse_llm_json(raw_response)
if not isinstance(parsed, dict):
    parsed = {}
```

## Regression test

`api/tests/test_qa_analysis_non_dict_response.py` drives `_run_whole_call_qa_analysis` with an array LLM response (mocking the LLM seams) and asserts empty results (`tags == []`, `summary == ""`, `score is None`) rather than an exception. The core parse-block behavior was verified directly: it raises `AttributeError` on the old code and degrades to empty defaults on the new code, while the normal dict path still maps correctly.

### DIFF
--- a/api/services/workflow/qa/analysis.py
+++ b/api/services/workflow/qa/analysis.py
@@ -210,6 +210,11 @@ async def run_per_node_qa_analysis(
             # top-level JSON array); coerce non-dict results so the .get()
             # lookups below don't raise AttributeError.
             if not isinstance(parsed, dict):
+                logger.warning(
+                    f"QA LLM returned non-object JSON for node '{node_name}' "
+                    f"on run {workflow_run_id}; got {type(parsed).__name__}, "
+                    "using empty QA result"
+                )
                 parsed = {}
             node_result["tags"] = parsed.get("tags", [])
             node_result["summary"] = parsed.get("summary", "")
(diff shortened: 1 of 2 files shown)


a6kme:
Thanks @Mubashirrrr for your contribution.

PR #421, opened 2026-06-07
feat: add Tone telephony provider (+91 TRAI-compliant via Exotel)

Bellatech-marketing:
Adds Tone (usetone.ai) as a telephony provider — TRAI-compliant +91 
numbers for AI voice agents, built on Exotel.

- New provider: `api/services/telephony/providers/tone/`
- Uses `ExotelFrameSerializer` from tuner-pipecat-sdk (WebSocket, 8kHz PCM)
- Implements full `TelephonyProvider` interface (matches Plivo structure)
- Added `TONE` to `WorkflowRunMode` enum

Tone API: https://api.usetone.ai/v1

### DIFF
--- a/api/enums.py
+++ b/api/enums.py
@@ -20,6 +20,7 @@ class CallType(Enum):
 class WorkflowRunMode(Enum):
     ARI = "ari"
     PLIVO = "plivo"
+    TONE = "tone"
     TWILIO = "twilio"
     VONAGE = "vonage"
     VOBIZ = "vobiz"
--- a/api/schemas/telephony_config.py
+++ b/api/schemas/telephony_config.py
@@ -28,6 +28,10 @@ from api.services.telephony.providers.telnyx.config import (
     TelnyxConfigurationRequest,
     TelnyxConfigurationResponse,
 )
+from api.services.telephony.providers.tone.config import (
+    ToneConfigurationRequest,
+    ToneConfigurationResponse,
+)
 from api.services.telephony.providers.twilio.config import (
     TwilioConfigurationRequest,
     TwilioConfigurationResponse,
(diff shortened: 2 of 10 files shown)


Bellatech-marketing:
@priteshkumar

chewwbaka:
Hi @Bellatech-marketing following are some blocking concerns in pr-

1. The Tone implementation says the Exotel App Bazaar WSS URL is configured statically in the dashboard, but Dograh’s generic telephony websocket URL requires dynamic per-call params: /api/v1/telephony/ws/{workflow_id}/{user_id}/{workflow_run_id}. How will Exotel/Tone supply those dynamic values for each call? If the dashboard URL is truly static, this likely needs a Tone-specific websocket route that resolves the workflow run from Exotel’s start payload/call id instead of using the generic dynamic URL.

2. We also need an end-to-end way to verify this provider. Signup/access to the Tone dashboard currently fails, so we can’t confirm the App Bazaar WSS setup, callback payloads, auth options, or whether dynamic run context can be passed correctly. Could you provide working test access or documented sample payloads/config steps?

3. Please add focused tests for the Tone provider flow: saving a Tone telephony config, initiating a call with the expected webhook URL, and handling/rejecting Tone callbacks correctly.

4. The PR adds ToneConfigurationRequest in api/services/telephony/providers/tone/config.py, but doesn’t add it to TelephonyConfigRequest in api/schemas/ telephony_config.py, so backend validation rejects provider: "tone" before it can be saved.

5. Why does ToneProvider.initiate_call() override the shared webhook_url with only workflow_run_id, while handle_tone_webhook() in tone/routes.py still requires workflow_id, user_id, and organization_id? This looks like Tone callbacks will 422 before the handler runs.

6. These Tone callback routes are public, but verify_inbound_signature() currently always returns True. Could we add real callback authentication here, such as a shared secret/API key in the callback URL or header, and reject requests before calling _process_status_update()?

chewwbaka:
Closing this because of blocking concerns, happy to reconsider if the above mentioned concerns are addressed in a new PR.

PR #422, opened 2026-06-08
feat: add Cartesia Sonic 3.5 as a TTS option

manasseh-zw:
## Summary
- add `sonic-3.5` to the Cartesia TTS model list
- make `sonic-3.5` the default Cartesia TTS model
- add a regression test covering the default model and service factory wiring

## Notes
- `sonic-3` remains available for compatibility
- this matches Cartesia's current stable base model identifier in the provider docs

## Testing
- not run locally in this repo because the local clone does not currently have a Python test environment configured


manasseh-zw:
Superseded by #423, which carries the same change on the correctly named  branch.

PR #423, opened 2026-06-08
feat: add Cartesia Sonic 3.5 as a TTS option

manasseh-zw:
## Summary
- add `sonic-3.5` to the Cartesia TTS model list
- make `sonic-3.5` the default Cartesia TTS model
- add a regression test covering the default model and service factory wiring

## Why
Cartesia now recommends `sonic-3.5` as the stable base model. Keeping `sonic-3` as the only default/visible option makes the OSS config lag behind the provider docs.

## Compatibility
- `sonic-3` remains available for compatibility


chewwbaka:
Thanks @manasseh-zw for this

PR #426, opened 2026-06-09
feat: add a start docker script

a6kme:
Adds a start docker script so that we have required values like OSS_JWT_SECRET at runtime


chewwbaka (approved):
LGTM

PR #431, opened 2026-06-12
fix: use system role for user-idle prompt injections

Mohamed-Mamdouh:
Fixes:
1) The UserIdleHandler injected its "are you still there?" and disconnect prompts as role="user" messages. These are agent-side directives, not user utterances, so they should be injected as role="system" to avoid polluting the conversation transcript with fake user turns and to read correctly by the LLM. Updated the realtime append tests to match.

2) forward ports 3000 (UI) and 8000 (API) in the devcontainer so the running services are reachable from the host.

3) Bump Tuner SDK version to 2.0.2

Testing and Validation:
Screenshot 2026-06-12 at 14 17 40

### DIFF
--- a/api/services/workflow/pipecat_engine_callbacks.py
+++ b/api/services/workflow/pipecat_engine_callbacks.py
@@ -46,14 +46,14 @@ class UserIdleHandler:
 
         if self._retry_count == 1:
             message = {
-                "role": "user",
+                "role": "system",
                 "content": "The user has been quiet. Politely and briefly ask if they're still there in the language that the user has been speaking so far.",
             }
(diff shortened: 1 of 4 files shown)


api/services/workflow/pipecat_engine_callbacks.py
```diff
@@ -46,14 +46,14 @@ async def handle_idle(self, aggregator):
 
         if self._retry_count == 1:
             message = {
-                "role": "user",
+                "role": "system",
```
a6kme, line 49: We cant do that since some models dont allow appending system prompt - example gemini models.

a6kme (changes requested):

PR #438, opened 2026-06-15
docs: add Resemble Detect and Signal tool prompts

devshahofficial:
## Summary
- add Dograh MCP prompts for creating Resemble HTTP tools
- include both Signal scoring and Detect media submission examples
- show how to attach those tools to a workflow before sensitive actions

## Validation
- `git diff --check`


a6kme:
Hey @devshahofficial - I don't believe its a good idea to add tool specific documentation in our official MCP docs. Please feel free to reach out to us at founders@dograh.com so that we can find better avenues for integration. Thanks!

PR #439, opened 2026-06-16
feat(telephony/vonage): wire up inbound calls

Samffprice:
## Summary

Vonage inbound calls currently fail with `Callee currently unavailable` even when the Application is configured per the [docs](https://docs.dograh.com/integrations/telephony/vonage). Five issues, all in `providers/vonage/provider.py` plus one supporting tweak to the inbound-route lookup, collectively break the flow:

1. **`can_handle_webhook` is a stub** returning `False` unconditionally — the unified `/api/v1/telephony/inbound/run` dispatcher's `_detect_provider` never identifies a Vonage payload. (Detected by `conversation_uuid` with `CON-` prefix, falling back to `region_url` containing `vonage.com`.)
2. **`generate_validation_error_response` is missing** — the validation-error path in the dispatcher hits `AttributeError`, falls back to generic TwiML, Vonage rejects as `no NCCOs available`.
3. **`parse_inbound_webhook` reads `direction` from a key Vonage's answer URL doesn't include** — the dispatcher then rejects with `Non-inbound call on /inbound/run:` and hangs up.
4. **`start_inbound_stream` is hardcoded** to a `talk` + `hangup` NCCO labeled `# Minimalist NCCO response for interface compliance` — even with the above three fixed, Vonage just answers and immediately hangs up.
5. **`call_uuid` lookup miss after a successful WebSocket connect** — outbound calls stamp `call_uuid` in `gathered_context`, but inbound calls go through `_create_inbound_workflow_run` which stores the same value under the generic `call_id` key. Fall back to `call_id`.

Plus one supporting change in `find_inbound_route_by_account` so the lookup tolerates providers (like Vonage) whose payload has no account-like identifier.

## What changed

| File | Change |
|---|---|
| `api/services/telephony/providers/vonage/provider.py` | Real `can_handle_webhook` detection; add `generate_validation_error_response`; default `direction` to `"inbound"` in `parse_inbound_webhook`; real `connect` NCCO in `start_inbound_stream`; `call_uuid` → `call_id` fallback when reading from `gathered_context` |
| `api/db/telephony_phone_number_client.py` | Relax `find_inbound_route_by_account` guard to `(provider, to_number)`; conditionally add the `account_id` WHERE clause. `uq_phone_numbers_org_address` already enforces the uniqueness needed for safety. |

Net diff: **+69 / -15** across two files. Twilio, Telnyx, Plivo, and other providers are unaffected — their `can_handle_webhook` overrides take precedence in `_detect_provider`, and they pass non-empty `account_id`.

## Verification

Tested end-to-end against a real Vonage account (Application + linked DID, signature verification on, NCCO answer URL pointing at this branch's `/inbound/run`):

| Stage | Before this PR | After this PR |
|---|---|---|
| Call rings the Vonage number | Vonage marks `FAILED — Callee currently unavailable` | Vonage marks `ANSWERED — The call was successful` |
| `/inbound/run` reaches Dograh | 200 but provider unrecognized; TwiML hangup returned | 200 with valid NCCO `connect` action to `wss://.../api/v1/telephony/ws/{wf}/{user}/{run}` |
| WebSocket | Never opened | Accepted; workflow run state set to `running`; bot greets caller |
| Status callbacks | n/a | Per-call event URL embedded in NCCO; downstream `started` / `answered` / `completed` flow as expected |

I'm running this on a self-hosted production deployment and inbound now answers cleanly and runs the workflow.

## Test plan

- [x] Real inbound call from external phone → number bound to a Vonage Application → workflow runs end-to-end
- [x] Existing Twilio inbound still works (regression check on the dispatcher's account_id path)
- [x] Existing Telnyx inbound still works
- [ ] CI: existing tests pass

## Notes

- I considered adding a unit test for `can_handle_webhook`, but the existing telephony tests for other providers focus on routes / signature verification rather than detection. Happy to add one if a reviewer prefers.
- The fallback from `call_uuid` → `call_id` could alternatively be solved by having the dispatcher write both keys. Doing it on the read side keeps the patch isolated to one provider; happy to flip if there's a preference for that approach.

### DIFF
--- a/api/db/telephony_phone_number_client.py
+++ b/api/db/telephony_phone_number_client.py
@@ -146,7 +146,13 @@ class TelephonyPhoneNumberClient(BaseDBClient):
         misses (e.g. legacy non-E.164 stored addresses); the caller should
         fall back to the fuzzy ``numbers_match`` path in that case.
         """
-        if not (provider and account_id_field and account_id and to_number):
+        # Allow ``account_id`` to be empty for providers whose inbound webhook
+        # payload doesn't carry an account-like identifier (e.g. Vonage's
+        # answer-URL POST contains only ``from`` / ``to`` / ``uuid`` /
+        # ``conversation_uuid``). When account_id is empty, match on
+        # ``(provider, address_normalized)``; ``uq_phone_numbers_org_address``
+        # guarantees that tuple resolves to at most one phone number per org.
+        if not (provider and to_number):
             return None
 
         normalized = normalize_telephony_address(to_number, country_hint=country_hint)
@@ -161,13 +167,16 @@ class TelephonyPhoneNumberClient(BaseDBClient):
                 )
                 .where(
                     TelephonyConfigurationModel.provider == provider,
-                    TelephonyConfigurationModel.credentials.op("->>")(account_id_field)
-                    == account_id,
                     TelephonyPhoneNumberModel.address_normalized
                     == normalized.canonical,
                     TelephonyPhoneNumberModel.is_active.is_(True),
                 )
             )
+            if account_id and account_id_field:
+                stmt = stmt.where(
+                    TelephonyConfigurationModel.credentials.op("->>")(account_id_field)
+                    == account_id
+                )
             if organization_id is not None:
                 stmt = stmt.where(
                     TelephonyConfigurationModel.organization_id == organization_id
--- a/api/routes/organization.py
+++ b/api/routes/organization.py
@@ -776,6 +776,10 @@ async def create_phone_number(
                 ),
             )
 
+    extra_metadata = dict(request.extra_metadata or {})
+    if request.smart_voicemail is not None:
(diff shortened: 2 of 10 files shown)


chewwbaka:
Hi @Samffprice , thanks for putting this together. I went through the PR, and since Vonage inbound calling is the main feature we want to prioritize and review first, I’ve opened a focused PR for that piece here: https://github.com/dograh-hq/dograh/pull/480

That PR keeps the scope limited to Vonage inbound support: signed webhook verification, inbound routing, NCCO websocket connect, event handling, tests, and docs.

This PR also includes additional functionality outside inbound calling, including smart voicemail / screen-and-forward support, phone-number extra_metadata schema changes, screening pipeline changes, Vonage call-control and transfer helpers, and UI controls for smart voicemail configuration. These may still be useful, but I think they should be handled in a separate PR so they can be reviewed independently.
Splitting these changes should make the review and merge path much cleaner. Thanks again for the work here.

PR #442, opened 2026-06-16
fix: add language field to CartesiaTTSConfiguration and pass to Cartesia TTS service

### LINKED ISSUE #432: [Bug]: TTS pipelines ignore Turkish language context (Cartesia and Default TTS pronounces Turkish with English phonetics)
### Describe the bug

Hi,

I am experiencing a critical issue with the TTS output when the LLM generates responses in Turkish. Even though the text is perfectly formatted in Turkish, the output audio pronounces the words using English phonetics.

For example, when the LLM says:
"Merhaba, ben Kadir. Size nasıl yardımcı olabilirim?"
The word "Size" is being pronounced as "sayz" (like the English word 'size') instead of the correct Turkish pronunciation.

 

### Steps to Reproduce

Cartesia: I tested with a specific Turkish Voice ID, but the behavior remained the same. It seems Dograh does not explicitly pass the language parameter/flag to the Cartesia API payload or send "en" language, causing Cartesia to fallback to English phonetics for text synthesis.

Dograh Default TTS: The exact same mispronunciation occurs using the default TTS provider built into Dograh.

The Dograh core audio/TTS pipeline seems to be hardcoded to an English locale (en or en-US) when sending payloads to external TTS providers, or it completely lacks dynamic language/locale detection based on the LLM's output language.

### Expected behavior

Dograh should either:

Automatically detect the text language and pass the correct language code to the TTS provider.

Or provide a language or locale configuration option in the assistant/agent settings (e.g., locale: "tr") so the TTS engine knows how to parse the phonetics correctly.

### Screenshots

Image

### Device details

_No response_

### Additional context

[cartesia_audio_2026-06-12T16_50_09+03_00.wav](https://github.com/user-attachments/files/28881825/cartesia_audio_2026-06-12T16_50_09%2B03_00.wav)
[20.wav](https://github.com/user-attachments/files/28881826/20.wav)

nuthalapativarun:
## Summary
- Adds a `language` field to `CartesiaTTSConfiguration` in `registry.py` (BCP-47 code, default `"en"`)
- Passes the language value into `CartesiaTTSSettings` when building the Cartesia TTS service in `service_factory.py`
- Adds two regression tests to `test_cartesia_tts_service_factory.py` confirming the default and non-English pass-through

## Why
`CartesiaTTSConfiguration` had no `language` field, so the pipecat `CartesiaTTSSettings.language` parameter was never set. Cartesia defaulted to English phonetics regardless of the conversation language, causing non-English speech (e.g. Turkish) to be synthesised with incorrect pronunciation.

## Changes
| File | Change |
|------|--------|
| `api/services/configuration/registry.py` | Added `language: str = Field(default="en", ...)` to `CartesiaTTSConfiguration` |
| `api/services/pipecat/service_factory.py` | Reads `user_config.tts.language` and passes it to `CartesiaTTSSettings(language=...)` |
| `api/tests/test_cartesia_tts_service_factory.py` | Added `test_cartesia_tts_configuration_default_language_is_english` and `test_create_cartesia_tts_service_passes_language_to_settings` |

## Testing
```bash
pytest api/tests/test_cartesia_tts_service_factory.py -v
# Before: CartesiaTTSSettings received no language → Cartesia defaulted to English
# After:  CartesiaTTSSettings.language="tr" when tts.language="tr" is configured
```

Closes #432


nuthalapativarun:
The `drift-check` failure was caused by `docs/api-reference/openapi.json` being out of sync with the FastAPI app — `main` gained a few new organization endpoints (billing/credits, mps-credits purchase-url, context) after this branch was created, so the checked-in spec snapshot was stale relative to the current route set. Regenerated it with `python -m scripts.dump_docs_openapi` and pushed the update. No application code changed.

a6kme:
Thanks @nuthalapativarun for your contribution.

PR #445, opened 2026-06-16
allow self-hosters to enable Stack Auth via Dockerfile build args (v33.0)

neggmmm:
## Problem

Self-hosters who want to use Stack Auth (Google / GitHub / social login)
instead of the built-in email-password flow cannot do so with the
current prebuilt `dograhai/dograh-ui` image, because `NEXT_PUBLIC_*`
environment variables are inlined into the browser bundle at
`npm run build` time by Next.js. Setting them as runtime env vars on the
container has no effect — the bundle was already compiled without them.

The only workaround today is to maintain a fork with a patched
Dockerfile, which creates unnecessary maintenance burden for anyone
deploying Dograh with Stack Auth.

## Solution

Add two optional `ARG` declarations to the **builder** stage of
`ui/Dockerfile`, immediately before the `npm run build` step:

- `NEXT_PUBLIC_STACK_PROJECT_ID`
- `NEXT_PUBLIC_STACK_PUBLISHABLE_CLIENT_KEY`

Each ARG is promoted to a matching `ENV` so Next.js picks it up during
the build. Self-hosters pass their project-specific values via
`docker build --build-arg` or their platform's build variable UI
(Railway, Coolify, etc.).

## Backward compatibility

- **When the args are not supplied** (the default), the values are empty
  strings and the UI behaves exactly as it does today — OSS / local auth,
  no Stack initialization. The official prebuilt image pipeline does not
  pass these args, so published images are unchanged.
- **When the args are supplied**, the Stack Auth client activates in the
  browser and the UI delegates authentication to the configured Stack
  project.
- No runtime code is modified — this is a build-config-only change.

## How to use

```bash
docker build \
  --build-arg NEXT_PUBLIC_STACK_PROJECT_ID=<your-project-id> \
  --build-arg NEXT_PUBLIC_STACK_PUBLISHABLE_CLIENT_KEY=<your-key> \
  -f ui/Dockerfile .
```

Or set them as build variables in Railway / Coolify / any Docker-based
PaaS — they are picked up automatically as Docker build args.

## Related context

Stack Auth (now Hexclave) is already fully supported on the backend via
the `AUTH_PROVIDER=stack` env var and the existing `stack_auth.py`
integration. This PR closes the gap on the frontend side for
self-hosted deployments.

### DIFF
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -3,27 +3,20 @@
 # Stage 1: Dependencies
 FROM node:20-alpine AS deps
 WORKDIR /app
-
 # Install Python and build dependencies for native modules
 # This helps with ARM64 builds and native module compilation
 RUN apk add --no-cache python3 make g++ libc6-compat
-
 # Copy package files
 COPY ui/package*.json ./
-
 # Clean install with proper handling of native modules
 RUN --mount=type=cache,target=/root/.npm npm ci
-
 # Stage 2: Builder
 FROM node:20-alpine AS builder
 WORKDIR /app
-
 # Install libc6-compat for native modules in builder stage too
 RUN apk add --no-cache libc6-compat
-
 # Copy dependencies from deps stage
 COPY --from=deps /app/node_modules ./node_modules
-
 # Copy all files needed for build
 COPY ui/package*.json ./
 COPY ui/tsconfig.json ./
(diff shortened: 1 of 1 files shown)


a6kme:
@neggmmm - Thanks for your contribution.

a6kme:
Hey @neggmmm - I have taken a different approach to enable this - which prevents an image rebuild. Please look at the documentation here https://docs.dograh.com/deployment/authentication

PR #446, opened 2026-06-16
fix(ui): release microphone stream on call teardown so a second test call works

Aymenbenpakiss:
## Problem

In the dashboard browser test call (**Call Voice Agent** → `useWebSocketRTC`), the call works the **first** time but a **second** attempt on the same page fails — the mic stays stuck / `Could not acquire media` — until the user reloads the page or clears site data and re-grants microphone permission. Reproduces across multiple browsers.

## Root cause

The `MediaStream` from `navigator.mediaDevices.getUserMedia()` is added to the `RTCPeerConnection` but never stored or explicitly stopped on teardown. On hangup, only `sender.track.stop()` runs (via `pc.getSenders()`); browsers can keep the microphone device held through the **original** `MediaStream` reference, so the next `getUserMedia()` is blocked. A page reload (what "clear cache" effectively does) drops the leaked stream — which is why the workaround works.

## Fix

Keep the stream in a `localStreamRef` and stop all of its tracks:
- in `cleanupConnection()` — the shared teardown path used by `stop()`, error handlers, and disconnects — right before closing the peer connection, and
- in the on-unmount cleanup `useEffect`.

This fully releases the microphone between calls. Minimal, additive change (14 lines); `sender.track.stop()` and the delayed `pc.close()` are left as-is (stopping an already-stopped track is a no-op).

## Verification

1. Open a workflow → Call Voice Agent, grant mic, talk briefly, hang up.
2. Without reloading or clearing anything, start a **second** call → connects, audio works, no re-permission prompt, no device-locked error.
3. The OS/tab mic-in-use indicator turns **off** after each hangup (proof the stream is released).

### DIFF
--- a/ui/public/embed/dograh-widget.js
+++ b/ui/public/embed/dograh-widget.js
@@ -633,6 +633,11 @@
       // Request microphone permission
       try {
         const stream = await navigator.mediaDevices.getUserMedia({ audio: true });
+        // Release any stream still held from a prior attempt before retaining
+        // the new one, so a re-entrant start can't leak the microphone.
+        if (state.stream) {
+          state.stream.getTracks().forEach(track => track.stop());
+        }
         state.stream = stream;
       } catch (micError) {
         // Handle specific microphone permission errors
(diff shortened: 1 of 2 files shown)


a6kme:
Thanks @Aymenbenpakiss for your contribution.

PR #453, opened 2026-06-19
feat: add docker-compose.dokploy.yaml for Dokploy / reverse-proxy deploys

tecnomanu:
## Summary

Adds a separate `docker-compose.dokploy.yaml` aimed at deployments behind a host-level reverse proxy (Dokploy's Traefik, but the same shape applies to any setup where TLS termination and public exposure are handled outside the stack — Coolify, Caprover, a plain Traefik/Nginx in front, etc.). The standard `docker-compose.yaml` is untouched.

The motivation: the stock compose assumes it owns the public ports (80/443) and ships with documented dev passwords for postgres/redis. Trying to deploy it on a platform that already runs a reverse proxy forces every operator to fork the file and edit the same handful of things — port bindings, password defaults, the `cloudflared` service, the lack of `host.docker.internal` mapping. This PR moves those deltas into a sibling file so the standard one stays clean and the Dokploy path is just "point the platform at `docker-compose.dokploy.yaml`."

## What it changes vs `docker-compose.yaml`

| Change | Why |
|---|---|
| `POSTGRES_PASSWORD`, `REDIS_PASSWORD`, `OSS_JWT_SECRET` are required (`:?`) | Safer for production deploys — no accidental ship with documented dev creds |
| `POSTGRES_USER` / `POSTGRES_DB` configurable via env | Supports migrating existing databases |
| postgres and redis no longer publish ports to the host | Only reachable via the Docker network — the platform proxy handles ingress |
| `api` and `ui` bind to `127.0.0.1` instead of `0.0.0.0` | Same reasoning — host proxy is the public entrypoint |
| nginx uses `expose: "80"` instead of `ports: 80/443` | TLS termination is the external proxy's job; routes 443 → nginx:80 internally |
| `cloudflared` service omitted | The host proxy already provides public exposure |
| api adds `extra_hosts: host.docker.internal:host-gateway` | Lets the ARI telephony provider reach an Asterisk instance on the host via `http://host.docker.internal:8088` |

Everything else (services, healthchecks, volumes, networks, posthog/telemetry env, MinIO setup, FORWARDED_ALLOW_IPS for proxied X-Forwarded-Proto) is preserved verbatim from the stock compose.

## Required env vars

The file header lists them with example generators:

```
POSTGRES_PASSWORD       openssl rand -hex 32
REDIS_PASSWORD          openssl rand -hex 32
OSS_JWT_SECRET          openssl rand -hex 64
PUBLIC_HOST             e.g. dograh.example.com
PUBLIC_BASE_URL         e.g. https://dograh.example.com
ENVIRONMENT             production
COMPOSE_PROFILES        remote
```

## Test plan

- [x] `docker compose -f docker-compose.dokploy.yaml config` validates with the required vars set.
- [x] `docker compose -f docker-compose.dokploy.yaml config` fails fast (with a clear message) when any required var is missing.
- [x] Running on a real Dokploy production stack (Asterisk 20.19 + PJSIP trunk + chan_websocket external media, ARI provider reaching Asterisk on the host via `host.docker.internal`) — stack comes up, API/UI healthy behind Traefik, MinIO reachable via the proxied public URL.

## Notes

- The standard `docker-compose.yaml` is not modified, so no existing user is affected.
- I'm running this in production from a setup at https://github.com/tecnomanu/dokploy-dograh/tree/test/ws-fix (also currently builds from the fork branch of #314 to apply the chan_websocket simple_bridge fix). That setup is the basis for this PR; the fix-related differences are not part of this PR.

### DIFF
--- a/docker-compose.dokploy.yaml
+++ b/docker-compose.dokploy.yaml
@@ -1,0 +1,287 @@
+# Dokploy-friendly compose for Dograh.
+#
+# Differences vs the standard docker-compose.yaml:
+#   - Postgres/Redis passwords are REQUIRED (no insecure defaults).
+#   - Postgres/Redis are not exposed on the host — only reachable inside the
+#     Docker network.
+#   - API and UI bind to 127.0.0.1 instead of 0.0.0.0; the public entrypoint
+#     is the reverse proxy that Dokploy (Traefik) runs in front of the stack.
+#   - nginx uses `expose` instead of `ports` for the same reason — Traefik
+#     handles TLS termination and routes 443 -> nginx:80 internally.
+#   - No `cloudflared` service. Dokploy already provides public exposure.
+#   - `host.docker.internal:host-gateway` is added to the API so the ARI
+#     telephony provider can reach an Asterisk instance running on the host
+#     (or another container reachable via the host network).
+#   - Postgres USER and DB are configurable via env to support migrating
+#     existing installations.
+#
+# Required environment variables (set these in Dokploy's env panel):
+#   POSTGRES_PASSWORD   # openssl rand -hex 32
+#   REDIS_PASSWORD      # openssl rand -hex 32
+#   OSS_JWT_SECRET      # openssl rand -hex 64
+#   PUBLIC_HOST         # e.g. dograh.example.com
+#   PUBLIC_BASE_URL     # e.g. https://dograh.example.com
+#   BACKEND_API_ENDPOINT
+#   MINIO_PUBLIC_ENDPOINT
+#   ENVIRONMENT=production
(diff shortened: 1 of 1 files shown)


a6kme:
Hey @tecnomanu - Thank you for your PR.

I would not recommend a separate docker compose file for this. We are already using a script to create more secure password for openssl for local setup https://github.com/dograh-hq/dograh/blob/main/scripts/start_docker.sh. Similarly, we generate secure password for Postgres from our remote setup script - https://github.com/dograh-hq/dograh/blob/main/scripts/setup_remote.sh 

We can extend it to create secure password for redis too. As of not exposing ports, this can be a system admin decision.

Issue #425, opened 2026-06-08
[Bug]: Describe the bug


hugocorreag79-ops:
### Describe the bug

After upgrading a working self-hosted Dograh deployment from **v1.33.0** to **v1.34.0 / latest**, voice test calls break completely because the browser cannot establish the signaling WebSocket connection (`wss://...`). 
This issue persists regardless of the domain layout (tried both separate subdomains and a unified single-domain proxy setup). The signaling WebSocket fails to connect or upgrade, throwing recurring connection error events in the browser console.

### Steps to Reproduce

1. Start with a working Dograh self-hosted deployment on **v1.33.0**.
2. Upgrade the Docker images (`dograh-ui` and `dograh-api`) to **v1.34.0** or **latest**.
3. Attempt to start a test voice call.
4. The signaling WebSocket fails to connect, producing connection errors in the browser console.

### Expected behavior

The signaling WebSocket connection should successfully open, upgrade, and connect (as it does on v1.33.0) to allow WebRTC calls to start.

### Device details

Ubuntu 24.04

### Additional context

Dograh Version: v1.34.0 / latest
Deployment: Docker Compose on VPS (Hostinger) KMV 8
Reverse Proxy: Nginx 
Browser: Google Chrome / Firefox

a6kme:
Hello @hugocorreag79-ops - We just pushed a fix for it in v1.37. Could you please give it another spin? Thanks!

PR #454, opened 2026-06-19
fix(ui): proxy WebSocket signaling upgrade so local web calls work (#425)

yogi6969:
## Problem

On local OSS deployments, starting a web call creates the run (`POST /workflow/{id}/runs → 200`) but the browser's signaling WebSocket to `/api/v1/ws/signaling/...` never connects, so WebRTC never negotiates. This is issue #425 — it blocks every web call on `>= 1.34.0`.

## Root cause

1.34.0 replaced the `next.config` `/api/* → BACKEND_URL` **rewrite** with the Route Handler `ui/src/app/api/v1/[...path]/route.ts`.

Route Handlers proxy HTTP fine (which is why `/api/v1/*` requests still return 200), but **they cannot upgrade WebSocket connections**. The removed *rewrite* used to carry that upgrade. Without it, the signaling socket has no proxy path and the connection fails before WebRTC can start.

`nginx` would proxy the upgrade, but it only runs in the `remote` compose profile — so local OSS deployments (browser → Next.js `:3010` directly) have nothing to carry it.

Note that `useWebSocketRTC.tsx` and `apiClient.ts` are unchanged between 1.33.0 and now — the WS URL has always been same-origin. The regression is purely the removed rewrite.

## Fix

Re-add a `beforeFiles` rewrite scoped to `/api/v1/ws/:path*` so the upgrade is proxied to the backend **before** the `[...path]` Route Handler can swallow it. `beforeFiles` runs ahead of filesystem routes/route handlers, so HTTP `/api/v1/*` is untouched and still flows through the Route Handler (auth/cookie handling intact).

## Verification

On a source build, the signaling WS now reports:

- **Server:** `WebSocket /api/v1/ws/signaling/... [accepted]` → `connection open` → `ICE completed` → `Registering audio data handler`
- **Client:** `WebSocket connected` → `Sent offer via WebSocket` → `ICE connection state: connected`

WebRTC negotiates end-to-end and the web call works again.

### DIFF
--- a/ui/next.config.ts
+++ b/ui/next.config.ts
@@ -8,20 +8,35 @@ const nextConfig: NextConfig = {
     serverSourceMaps: true,
   },
   async rewrites() {
-    return [
-      {
-        source: "/ingest/static/:path*",
-        destination: "https://us-assets.i.posthog.com/static/:path*",
-      },
-      {
-        source: "/ingest/:path*",
-        destination: "https://us.i.posthog.com/:path*",
-      },
-      {
-        source: "/ingest/decide",
-        destination: "https://us.i.posthog.com/decide",
-      },
-    ];
+    return {
+      // beforeFiles runs before Next.js route handlers, so the WebSocket
+      // signaling path is proxied straight to the backend instead of being
+      // swallowed by api/v1/[...path]/route.ts — a Route Handler that proxies
+      // HTTP fine but CANNOT upgrade WebSocket connections. The pre-1.34.0
+      // /api proxy *rewrite* used to carry this upgrade; removing it broke all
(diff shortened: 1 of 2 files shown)


a6kme:
Hey @yogi6969 - thanks for your contribution. I have taken a different approach because the rewrite was a change that I needed to test more thoroughly because we support multiple deployment options behind reverse proxies. 

This issue lived for so long because it was on a local Docker deployment and I did not test it properly. 😥

I have pushed a heuristic-based approach to unblock the local deployment mode, and I will give it a bit more thought on how to handle this URL configuration across all of our deployment options, including remote deployment, custom domain deployment, local deployment, development setup, Kubernetes deployment to name a few.

PR #458, opened 2026-06-19
feat(scripts): generate REDIS_PASSWORD on setup, plumb through compose

tecnomanu:
## Summary

Follow-up to the feedback on #453 — instead of a separate compose file, extend the existing setup scripts to generate a secure `REDIS_PASSWORD`, and update the compose files to consume it. As suggested by @a6kme:

> We can extend it to create secure password for redis too.

Backwards-compatible: existing installs that haven't regenerated their `.env` keep the old `redissecret` value via a fallback, so nothing breaks on upgrade.

## What changes

| File | Change |
|---|---|
| [scripts/setup_remote.sh](https://github.com/dograh-hq/dograh/blob/main/scripts/setup_remote.sh) | Generate `REDIS_PASSWORD=$(openssl rand -hex 32)` alongside the existing `OSS_JWT_SECRET` / `POSTGRES_PASSWORD` generation and write it to the rendered `.env` |
| [scripts/start_docker.sh](https://github.com/dograh-hq/dograh/blob/main/scripts/start_docker.sh) | Mirror the existing `OSS_JWT_SECRET` block: if `REDIS_PASSWORD` is missing from `.env`, generate one via `generate_secret` (which falls back through python3 → openssl → /dev/urandom) |
| [docker-compose.yaml](https://github.com/dograh-hq/dograh/blob/main/docker-compose.yaml) | Replace 3 literal `redissecret` occurrences with `${REDIS_PASSWORD:-redissecret}`: redis `--requirepass`, redis healthcheck, and API `REDIS_URL` |
| [docker-compose-local.yaml](https://github.com/dograh-hq/dograh/blob/main/docker-compose-local.yaml) | Replace 2 literal `redissecret` occurrences with `${REDIS_PASSWORD:-redissecret}`: redis `--requirepass` and healthcheck |

The `:-redissecret` fallback is intentional — it preserves backwards compatibility for users whose `.env` predates this change. They keep working with the old value until they choose to regenerate. New installs through either script get a secure random hex.

## Test plan

- [x] `docker compose -f docker-compose.yaml config` validates and renders `REDIS_URL: redis://:redissecret@redis:6379` when `REDIS_PASSWORD` is unset (backwards compat).
- [x] Same command with `REDIS_PASSWORD=abc123` exported renders `REDIS_URL: redis://:abc123@redis:6379` (env var honored).
- [x] `docker-compose-local.yaml` validates analogously.
- [x] `bash -n scripts/setup_remote.sh` and `bash -n scripts/start_docker.sh` — both syntax-clean.
- [x] `scripts/start_docker.sh` on a clean directory creates both `OSS_JWT_SECRET` and `REDIS_PASSWORD` entries in `.env`; re-running is idempotent (existing values preserved, no duplicates).

## Notes

- Scope is intentionally minimal — only what was explicitly invited in #453's feedback. The other items from that PR (host binding choices, `host.docker.internal` for ARI, etc.) are not included here.
- Did not touch the PowerShell counterparts (`start_docker.ps1`, etc.) — happy to add them in this PR if preferred, or leave for a separate one.

### DIFF
--- a/docker-compose-local.yaml
+++ b/docker-compose-local.yaml
@@ -27,11 +27,11 @@ services:
     ports:
       - "6379:6379"
     command: >
-      --requirepass ${REDIS_PASSWORD:-redissecret}
+      --requirepass redissecret
     volumes:
       - redis_data:/data
     healthcheck:
-      test: ["CMD", "redis-cli", "-a", "${REDIS_PASSWORD:-redissecret}", "ping"]
+      test: ["CMD", "redis-cli", "-a", "redissecret", "ping"]
       interval: 3s
       timeout: 10s
       retries: 10
--- a/docker-compose-local.yaml
+++ b/docker-compose-local.yaml
@@ -27,11 +27,11 @@ services:
     ports:
       - "6379:6379"
     command: >
-      --requirepass redissecret
+      --requirepass ${REDIS_PASSWORD:-redissecret}
     volumes:
       - redis_data:/data
     healthcheck:
-      test: ["CMD", "redis-cli", "-a", "redissecret", "ping"]
+      test: ["CMD", "redis-cli", "-a", "${REDIS_PASSWORD:-redissecret}", "ping"]
       interval: 3s
       timeout: 10s
       retries: 10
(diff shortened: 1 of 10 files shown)


a6kme:
Thanks @tecnomanu for your contribution 🚀

PR #461, opened 2026-06-22
feat(storage): support custom S3 endpoint, signature version, and addressing style

skymoore:
## Summary

Adds optional overrides so the existing `S3FileSystem` storage backend can target any **S3-compatible server** (MinIO, [rustfs](https://github.com/rustfs/rustfs), Ceph, ...), not just AWS — while keeping the bucket **private** by issuing real presigned URLs.

Three new (all optional) environment variables:

| Variable | Default | Description |
|---|---|---|
| `S3_ENDPOINT_URL` | unset | Custom S3 endpoint, e.g. `https://s3.example.com`. Unset → AWS default resolution. |
| `S3_SIGNATURE_VERSION` | unset | botocore signature version. Unset → botocore default; set `s3v4` for servers that require SigV4. |
| `S3_ADDRESSING_STYLE` | unset | `auto` (default), `path`, or `virtual`. Many S3-compatible servers / TLS setups require `path`. |

Credentials continue to come from the standard `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`.

## Motivation

The `MinioFileSystem` backend does not sign URLs — it returns plain unsigned object URLs and relies on the bucket being anonymously public-readable. That forces a private bucket to be made public. The `S3FileSystem` backend already issues real presigned URLs, but it hard-coded `aioboto3.Session()` with no `Config` and no `endpoint_url`, so it could only talk to AWS and defaulted to SigV2 presigning — which non-AWS servers (e.g. rustfs) reject with `SignatureDoesNotMatch`.

With these overrides you can point the S3 backend at a self-hosted S3 server, sign with SigV4 + path-style, and keep the bucket private.

## Changes

- `api/constants.py` — add `S3_ENDPOINT_URL`, `S3_SIGNATURE_VERSION`, `S3_ADDRESSING_STYLE`.
- `api/services/filesystem/s3.py` — accept the overrides (constructor args defaulting to the env constants); build a `botocore.Config` **only** when an override is set; pass `endpoint_url`/`config` to every `session.client("s3", ...)` call via a shared `_client_kwargs()` helper.
- Docs: `docs/developer/environment-variables.mdx`, `api/.env.example`, `docker-compose.yaml`.

## Backwards compatibility

Defaults are unchanged. When none of the new vars are set, no `Config` is built and `endpoint_url` is omitted, so client construction is byte-for-byte identical to before. Existing AWS S3 and MinIO deployments are unaffected.

## Notes

- Presigned URLs point at `S3_ENDPOINT_URL`, so that host must be browser-reachable.
- Browsers fetch transcripts cross-origin, so the bucket needs a CORS rule (`GET`/`HEAD`) for the app origin — configured on the storage server (`PutBucketCors`), not in Dograh. This is documented in the env-vars page.

### DIFF
--- a/api/.env.example
+++ b/api/.env.example
@@ -18,6 +18,13 @@ ENABLE_AWS_S3="false"
 # AWS_SECRET_ACCESS_KEY=""
 # S3_BUCKET=""
 # S3_REGION=""
+# --- S3-compatible servers (MinIO, rustfs, Ceph, ...) ---
+# Use the S3 backend (ENABLE_AWS_S3=true) against a non-AWS, S3-compatible
+# server by overriding the endpoint and signing. Unlike the MinIO backend, the
+# S3 backend emits real presigned URLs, so the bucket can stay private.
+# S3_ENDPOINT_URL=""        # e.g. https://s3.example.com (blank = AWS default)
+# S3_SIGNATURE_VERSION=""   # blank = botocore default; set "s3v4" if the server requires SigV4
+# S3_ADDRESSING_STYLE=""    # blank = auto; set "path" if the server / TLS cert requires path-style
 
 # MinIO Configuration if using containerised MinIO instead of
 # AWS S3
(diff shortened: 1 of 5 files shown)


a6kme:
Thanks @skymoore for your contribution 🚀 

I will be pushing small cleanups on top of this in main.

PR #464, opened 2026-06-23
Ability to pass data from the voice widget

coxfrederic:
Adding ability to pass custom variables to the call context, for example passing a username if the caller is known from inside an ERP system for example.


a6kme:
Hello @coxfrederic - Thank you for your PR. I will be taking this up with a different approach sometime early next week.

Issue #583, opened 2026-07-28
[Bug]: End-call tool customMessage is silently dropped (but still logged to transcript) in realtime/speech-to-speech mode


skyfrostmage:
### Describe the bug

When `is_realtime` is enabled (any speech-to-speech provider), an end-call tool configured with `messageType: "custom"` / `customMessage` never plays its audio. The message **is** written to the run transcript (via `persist_to_logs=True`), so it looks delivered when reviewing logs, but the caller hears nothing. In cascading (non-realtime) mode the same tool works correctly.

## Environment

- `is_realtime: true`
- Reproduced on grok realtime (`grok-voice-think-fast-1.0`)
- Same code path affects OpenAI Realtime, Azure Realtime, Gemini Live, and Ultravox

### Steps to Reproduce

1. Create an `end_call` tool with `messageType: "custom"` and a `customMessage`.
2. Attach it to a node in a workflow whose model config has `is_realtime: true`.
3. Run a call and reach the point where the agent ends via that tool.

## Expected

The `customMessage` is spoken before the call ends.

## Actual

No audio for the message. It still appears in the transcript as an assistant/tool line, so it looks delivered. The caller hears only whatever the model generated on its own (often a flat "I'll close things out" instead of the configured message).

## Root cause

- The realtime services intercept `TTSSpeakFrame` and only honor the **first** one (the greeting); every subsequent `TTSSpeakFrame` is dropped. Example, `grok_realtime.py`: *"TTSSpeakFrame after initial context already handled — Grok Realtime owns audio generation, ignoring."* The same pattern exists in `azure_realtime.py`, `openai_realtime.py`, `gemini_live.py`, and `ultravox_realtime.py`.
- End-call custom messages are queued as `TTSSpeakFrame(customMessage, persist_to_logs=True)` (see `pipecat_engine_custom_tools.py`), so they hit exactly that dropped path while still being logged.

The greeting already proves realtime models can speak fixed text on command — `static_greeting.py` / `_handle_initial_greeting` injects a "say the following exactly, word-for-word" conversation item and forces a response. The same mechanism could be applied to end-call (and transition) `TTSSpeakFrame`s in realtime mode instead of dropping them.

amaanJvd:
[...]

PR #592, opened 2026-07-30
fix(realtime): speak end-call/transition messages on OpenAI Realtime (#583) [PoC]

amaanJvd:
> **Scope / status:** proof-of-concept for **OpenAI Realtime only**. The routing is unit-tested, but the live audio + call-end timing can only be validated on a real speech-to-speech call — I have not been able to run that. Opening it as a reviewable first cut; happy to fan out to the other providers once the approach is confirmed and someone can verify on a live provider. See my analysis in #583.

## Problem
With `is_realtime: true`, an end-call tool configured with `messageType: "custom"` never plays its `customMessage` — the caller hears nothing, yet the message is written to the transcript (via `persist_to_logs=True`) so it looks delivered. Cascading (non-realtime) mode works fine.

## Root cause
Each realtime provider special-cases the **first** `TTSSpeakFrame` as the greeting (routing it through `_handle_initial_greeting` → `format_static_greeting_prompt`, which makes the model say the exact text) and then **drops every subsequent one** with an "owns audio generation, ignoring" log. End-call/transition messages are queued as later `TTSSpeakFrame`s, so they hit that dropped path.

## Fix (this PR)
In `DograhOpenAIRealtimeLLMService.process_frame`, post-greeting `TTSSpeakFrame`s are now spoken verbatim via a forced response (`_send_manual_response_create(instructions=format_say_verbatim_prompt(text), tool_choice="none")`) instead of being dropped. Adds `format_say_verbatim_prompt` (a greeting-agnostic variant).

Timing is already handled upstream: `_create_end_call_handler` ends the call with `abort_immediately=False` whenever a message is queued, so the existing "wait until the bot stops speaking" path hangs up only after this response finishes.

## Tests
`tests/test_openai_realtime_end_call_message.py` — a post-greeting message triggers a verbatim response; empty frames are ignored; messages before session-ready are skipped, not crashed. (Routing only — live audio still needs a real-call check.)

## Not in this PR
The same drop exists in `grok_realtime`, `azure_realtime`, `gemini_live`, and `ultravox_realtime`, each with its own response API — intentionally left for follow-up so this stays small and reviewable.

Refs #583

---
## Summary by cubic
Speak end‑call and node‑transition messages during OpenAI Realtime calls instead of dropping them. This prevents silent hangups where the transcript showed a message was delivered.

- **Bug Fixes**
  - Post‑greeting `TTSSpeakFrame`s now trigger a manual response (`tool_choice: "none"`) using `format_say_verbatim_prompt(...)` so the model says the exact text.
  - Added `format_say_verbatim_prompt` for mid‑call fixed lines.
  - Empty frames are ignored; messages before session‑ready are skipped with a warning.
  - Hangup timing unchanged; the call ends after the spoken message finishes.
  - Scoped to OpenAI Realtime; unit tests cover the routing.

<sup>Written for commit c0b3c03a1cf5e9e5bbc497948db4e8e033957788. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

Adds OpenAI Realtime support for speaking fixed post-greeting messages such as node-transition lines and end-call goodbyes. Execution against the frame router reproduced two delivery failures: a message arriving while the initial greeting waits for session readiness is discarded, and back-to-back messages create overlapping Realtime responses so a later message can be rejected. Do not merge until fixed speech is retained through session initialization and serialized across response completion.

### T-Rex validation blocked
The repository's focused pytest command could not collect in the reconstructed runtime because the `dotenv` package was missing. The dedicated frame-routing harness completed successfully and reproduced both reported failures. [Configure VMs](https://app.greptile.com/-/settings/trex#environments)

<h3>Confidence Score: 3/5</h3>

Not safe to merge because the new fixed-speech path can drop caller-facing messages in normal session-start and multi-message flows.

Two independent failures were reproduced in the core behavior added by this change: one drops speech before session readiness, and the other sends overlapping responses instead of preserving message order.

**Files Needing Attention:** `api/services/pipecat/realtime/openai_realtime.py` needs deferred-message handling and response serialization; the focused tests should cover both timing paths.

**What T-Rex did**
- T-Rex produced proofs for a posted P1 finding, including a focused validation script and harness output that demonstrate the race condition in the TTSSpeakFrame race.
- T-Rex performed a general contract validation, running the reproduced command and inspecting harness outputs that show the fixed message was lost and how the OpenAI error handling behaves.

<a href="https://app.greptile.com/trex/runs/16525832/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. General comment 

   <a href="#">P1</a> **Post-greeting speech is dropped while the initial greeting waits for session readiness**

   - **Bug**
     - A greeting TTSSpeakFrame sets `_handled_initial_context=True` and stores `_pending_initial_greeting_text`. If another fixed TTSSpeakFrame arrives before `_api_session_ready`, `process_frame` routes it to `_speak_verbatim`, which explicitly logs `skipping` and returns. When the session updates, only the greeting is sent; the fixed message is neither queued nor retried.
   - **Cause**
     - `_speak_verbatim` has no deferred-message queue or pending state for the not-ready branch, even though the greeting path has `_pending_initial_greeting_text` handling.
   - **Fix**
     - Queue post-greeting fixed texts received before readiness and drain them only after the initial greeting response has completed (or serialize them through a single response scheduler). Do not clear/drop the queue on `session.updated`.

2. General comment 

   <a href="#">P1</a> **Back-to-back fixed TTSSpeakFrames issue concurrent OpenAI Realtime response.create requests**

   - **Bug**
     - With `_handled_initial_context=True` and `_api_session_ready=True`, two fixed TTSSpeakFrames synchronously invoke `_send_manual_response_create` twice before any response completion. The produced events target the default conversation and have no `conversation='none'` setting. OpenAI permits only one active response for that conversation, so the second create is rejected with `conversation_already_has_active_response`, losing the second fixed message.
   - **Cause**
     - `_speak_verbatim` has no active-response tracking, response-completion hook, queue, cancellation behavior, or locking; `_send_manual_response_create` always sends a new `ResponseCreateEvent`.
   - **Fix**
     - Serialize fixed messages: maintain a FIFO queue and active forced-response state, send the next item only after `response.done` for the current forced response, and clear/reconcile it on cancellation/disconnect. Alternatively explicitly cancel and replace only where interruption is desired.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(realtime): speak end-call/transition..."](https://github.com/dograh-hq/dograh/commit/c0b3c03a1cf5e9e5bbc497948db4e8e033957788) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48579968)</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/api/services/pipecat/realtime/openai_realtime.py
+++ b/api/services/pipecat/realtime/openai_realtime.py
@@ -25,7 +25,10 @@ from typing import Any
 
 from loguru import logger
 
-from api.services.pipecat.realtime.static_greeting import format_static_greeting_prompt
+from api.services.pipecat.realtime.static_greeting import (
+    format_say_verbatim_prompt,
+    format_static_greeting_prompt,
+)
 from pipecat.frames.frames import (
     BotStartedSpeakingFrame,
     BotStoppedSpeakingFrame,
@@ -89,10 +92,20 @@ class DograhOpenAIRealtimeLLMService(OpenAIRealtimeLLMService):
                 else:
                     await self._handle_context(self._context)
             else:
-                logger.warning(
-                    f"{self}: TTSSpeakFrame after initial context already "
-                    "handled — OpenAI Realtime owns audio generation, ignoring"
-                )
+                # Fixed-text messages queued after the greeting — end-call
+                # goodbye messages and node-transition lines — must still be
+                # spoken. The realtime model owns audio generation, so we can't
+                # route them through TTS; instead ask the model to say the text
+                # verbatim. Previously these frames were dropped, so the caller
+                # heard nothing even though the message was still written to the
+                # transcript (#583).
+                text = frame.text.strip() if frame.text else ""
+                if text:
+                    await self._speak_verbatim(text)
+                else:
+                    logger.debug(
+                        f"{self}: empty TTSSpeakFrame after initial context, ignoring"
+                    )
             # Don't forward the frame; the audio path is owned by the realtime
             # service itself.
             return
@@ -157,6 +170,31 @@ class DograhOpenAIRealtimeLLMService(OpenAIRealtimeLLMService):
         self._handled_initial_context = True
         self._context = context
         await self._create_initial_greeting_response(greeting_text)
+
+    async def _speak_verbatim(self, text: str):
+        """Have the realtime model say a fixed mid-call line (end-call goodbye
+        or node-transition message) out loud, since these can't go through TTS.
+
+        The end-call flow ends the call with ``abort_immediately=False`` when a
+        message is queued, so the existing wait-for-bot-to-stop-speaking path
+        handles hanging up only after this response finishes.
+        """
+        if self._disconnecting:
+            return
+        if not self._api_session_ready:
+            # No deferral queue for post-greeting messages; a message that
+            # arrives before the session is ready is rare (the greeting has
+            # already completed by then). Log rather than silently drop.
+            logger.warning(
+                f"{self}: verbatim message received before session ready, "
+                f"skipping: {text!r}"
+            )
+            return
+        await self._ensure_conversation_setup()
+        await self._send_manual_response_create(
+            instructions=format_say_verbatim_prompt(text),
+            tool_choice="none",
+        )
 
     async def _create_initial_greeting_response(self, greeting_text: str):
         if self._disconnecting:
--- a/api/services/pipecat/realtime/static_greeting.py
+++ b/api/services/pipecat/realtime/static_greeting.py
@@ -6,3 +6,15 @@ def format_static_greeting_prompt(greeting_text: str) -> str:
         "caller to respond. Do not add anything before or after it.\n\n"
         f'"{greeting_text}"'
     )
+
+
+def format_say_verbatim_prompt(text: str) -> str:
+    """Instruction for a mid-call fixed message (e.g. an end-call goodbye or a
+    node-transition line): say the exact text and nothing else. Unlike the
+    greeting prompt, it makes no assumptions about the call having just
+    connected."""
+    return (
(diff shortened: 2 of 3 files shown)


a6kme:
Hey @amaanJvd - I thought about it, but was not convinced that this is the right approach. There is just so many edge cases that can come and I would have to spend more time thinking about it. 

For now, this can be handled by prompting the LLM to say anything specific before calling the EndCall tool. I will definitely get back to this in future when I learn more about the RealTime APIs.

amaanJvd:
Totally fair — thanks for taking the time to think it through, and no worries at all. Agreed there are a lot of realtime edge cases; the prompt-the-LLM-before-EndCall workaround makes sense as the pragmatic path for now. Happy to revisit alongside you if you dig into the Realtime APIs later. Appreciate the consideration!

PR #595, opened 2026-07-30
feat(telephony): add VoxPro (India VNO) managed-carrier provider

AManjunath-Voxpro:
VoxPro is a licensed Indian VNO offering managed PSTN (own 1400/1600 range) via a connector REST API — Twilio-style simplicity, India-native, DPDP-compliant.

Unlike the self-hosted **ARI** provider, the Dograh user never exposes their own Asterisk: they authenticate with a **per-tenant API key** and VoxPro originates/bridges the call on its own carrier + trunk, streaming **Plivo/Twilio-standard mu-law** audio to Dograh over the WebSocket.

### What this adds
- `api/services/telephony/providers/voxpro/{config,provider,transport,serializers,__init__}.py`
- Registration: one line in `providers/__init__.py`, `WorkflowRunMode.VOXPRO`, and the `TelephonyConfigRequest/Response` union in `schemas/telephony_config.py`
- Unit tests: `api/tests/telephony/voxpro/test_provider.py` (7 passing)

### Design notes
- **Managed carrier** (like `vobiz`), **call-control** for control (like `telnyx`): `initiate_call`/`transfer`/`hangup`/`status` are REST to the VoxPro connector; media is the Plivo-compatible WebSocket.
- **Reuses the published `PlivoFrameSerializer`** — no pipecat change needed.
- Isolation: a tenant key can only originate on its own trunk/DIDs (no global ARI exposure).

### Config fields
`api_key` (per-tenant, sensitive), `tenant_id` (X-Tenant-ID), `api_base`, `from_numbers` (DIDs).

Happy to provide **sandbox VoxPro credentials** for CI, and to iterate on conventions. Maintained by VoxPro Solutions.

> Note: opened a separate issue about the published `tuner-pipecat-sdk` missing a few modules `main` imports (blocks `pip`-only test collection) — final CI here should run in the Docker dev env.

---
## Summary by cubic
Adds the `voxpro` managed telephony provider for India with REST call control and a Plivo-compatible media WebSocket. Secures transfer completion with authenticated callbacks and durable, idempotent completion events.

- **New Features**
  - Registered `voxpro` (`WorkflowRunMode.VOXPRO`) with `VoxProConfigurationRequest/Response`, UI metadata, and `X-Tenant-ID` scoping.
  - Outbound + media: originate via the connector; stream over `/api/v1/telephony/ws/{workflow_id}/{organization_id}/{workflow_run_id}` using the Plivo serializer at 8 kHz mu-law with `auto_hang_up=false`.
  - Webhooks and numbers: HMAC-SHA256 (`X-VoxPro-Signature`) verification using the tenant API key; phone-number ownership checks via the connector inventory.

- **Bug Fixes**
  - Transfers are durable and idempotent: persist terminal events before publish; waiter reads stored result; clear on context removal; never skip publish on persistence errors.
  - Correct transfer resolution: persist `call_id` into `gathered_context` (outbound via `provider_metadata`, inbound via WS start); mounted WS route; `WEBHOOK_ENDPOINT=None`; clearer 400s via `generate_validation_error_response`.
  - Transfer-result endpoint is authenticated: verify `X-VoxPro-Signature` resolved via transfer context/run before claiming; reject unknown/unsigned/wrong-key callbacks; idempotent via `claim_transfer_step`.

<sup>Written for commit 175a1569e446430ce324e3317464910ab96a9a4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

Adds VoxPro as a managed telephony provider, including configuration, outbound call control, inbound media handling, phone-number validation, and transfer-result callbacks. It also preserves terminal transfer events so a callback received before the workflow subscribes can be recovered.

A runtime check of the transfer-result flow could not reach its callback assertions because importing the full telephony router fails before the route is exercised. The affected callback path should be re-run once the application runtime can import all configured provider dependencies.

## T-Rex validation blocked
The focused callback harness was blocked by a missing Python package: `azure`. Importing the telephony router loads Pipecat's Azure STT integration, which raises `ImportError: Missing module: No module named 'azure'` before the callback request, duplicate suppression, or durable-result recovery can execute. [Configure VMs](https://app.greptile.com/-/settings/trex#environments)

<h3>Confidence Score: 4/5</h3>

The new VoxPro transfer callback is a core call-completion path and requires a successful runtime check before it can be considered ready to merge.

The transfer callback behavior could not be exercised end to end. The available execution stopped during telephony-router import, leaving the callback's authentication, duplicate handling, and pre-subscription result recovery unverified.

**Files Needing Attention:** api/services/telephony/providers/voxpro/routes.py and the application dependency configuration that imports Pipecat Azure STT during telephony-router startup.

**What T-Rex did**
- T-Rex attempted to validate the transfer callback path by running the VoxPro transfer-callback harness with a recreated runtime, an in-memory Redis fake, the actual CallTransferManager, the VoxPro route, and the HMAC verifier.
- The run failed during imports because the full telephony router could not be loaded; Pipecat's Azure STT integration could not import the azure package.
- Earlier attempts were blocked by the dangling .trex-venv interpreter and missing dependencies; the latest available capture still fails while importing the full router.
- T-Rex produced proof for a posted P1 finding.

</details>

<sub>Reviews (9): Last reviewed commit: ["fix(voxpro): authenticate the transfer-r..."](https://github.com/dograh-hq/dograh/commit/175a1569e446430ce324e3317464910ab96a9a4b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48656250)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/api/enums.py
+++ b/api/enums.py
@@ -51,6 +51,7 @@ class WorkflowRunMode(Enum):
     VOBIZ = "vobiz"
     CLOUDONIX = "cloudonix"
     TELNYX = "telnyx"
+    VOXPRO = "voxpro"
     WEBRTC = "webrtc"
     SMALLWEBRTC = "smallwebrtc"
     TEXTCHAT = "textchat"
--- a/api/schemas/telephony_config.py
+++ b/api/schemas/telephony_config.py
@@ -40,6 +40,10 @@ from api.services.telephony.providers.vonage.config import (
     VonageConfigurationRequest,
     VonageConfigurationResponse,
 )
+from api.services.telephony.providers.voxpro.config import (
+    VoxProConfigurationRequest,
+    VoxProConfigurationResponse,
+)
 
 # Discriminated union for incoming save requests. Pydantic dispatches on the
 # ``provider`` Literal field of each request class. Replaces the manual
@@ -53,6 +57,7 @@ TelephonyConfigRequest = Annotated[
         TwilioConfigurationRequest,
         VobizConfigurationRequest,
         VonageConfigurationRequest,
+        VoxProConfigurationRequest,
     ],
     Field(discriminator="provider"),
 ]
@@ -73,6 +78,7 @@ class TelephonyConfigurationResponse(BaseModel):
     cloudonix: Optional[CloudonixConfigurationResponse] = None
     ari: Optional[ARIConfigurationResponse] = None
     telnyx: Optional[TelnyxConfigurationResponse] = None
+    voxpro: Optional[VoxProConfigurationResponse] = None
 
 
 # ---------------------------------------------------------------------------
--- a/api/services/telephony/call_transfer_manager.py
+++ b/api/services/telephony/call_transfer_manager.py
@@ -98,6 +98,7 @@ class CallTransferManager:
             redis = await self._get_redis()
             context = await self.get_transfer_context(transfer_id)
             key = TransferRedisChannels.transfer_context_key(transfer_id)
+            await redis.delete(TransferRedisChannels.transfer_result_key(transfer_id))
             if context and context.original_call_sid:
                 index_key = TransferRedisChannels.transfer_context_by_call_sid_key(
                     context.original_call_sid
@@ -144,7 +145,23 @@ class CallTransferManager:
 
             redis = await self._get_redis()
             channel = TransferRedisChannels.transfer_events(event.transfer_id)
-            await redis.publish(channel, event.to_json())
+            payload = event.to_json()
+            # Persist terminal outcomes BEFORE publishing. Redis pub/sub keeps
+            # nothing for a subscriber that has not arrived yet, and a fast
+            # provider callback can land before the workflow starts waiting —
+            # the event would be lost and the transfer would time out even
+            # though the destination answered. The waiter reads this key right
+            # after subscribing, which closes that window from both sides.
+            if event.type in (
+                TransferEventType.DESTINATION_ANSWERED,
+                TransferEventType.TRANSFER_FAILED,
+            ):
+                await redis.setex(
+                    TransferRedisChannels.transfer_result_key(event.transfer_id),
+                    300,
+                    payload,
+                )
+            await redis.publish(channel, payload)
             logger.info(f"Published {event.type} event for {event.transfer_id}")
         except Exception as e:
             logger.error(f"Failed to publish transfer event: {e}")
@@ -170,6 +187,24 @@ class CallTransferManager:
             logger.info(
                 f"Waiting for transfer completion on {channel} (timeout: {timeout_seconds}s)"
             )
+
+            # Subscribe first, then read the durable copy: anything published
+            # before we subscribed is still here, and anything published after
+            # arrives on the subscription. Without this, a sub-second callback
+            # (the connector-driven providers are that fast) is simply missed.
+            stored = await redis.get(
+                TransferRedisChannels.transfer_result_key(transfer_id)
+            )
+            if stored:
+                try:
+                    event = TransferEvent.from_json(stored)
+                    logger.info(
+                        f"Read stored {event.type} event for {transfer_id} "
+                        "(published before the wait subscribed)"
+                    )
+                    return event
+                except Exception as e:
+                    logger.error(f"Failed to parse stored transfer event: {e}")
 
             # Wait for completion event with timeout
             async def wait_for_message():
--- a/api/services/telephony/call_transfer_manager.py
+++ b/api/services/telephony/call_transfer_manager.py
@@ -156,11 +156,22 @@ class CallTransferManager:
                 TransferEventType.DESTINATION_ANSWERED,
                 TransferEventType.TRANSFER_FAILED,
             ):
-                await redis.setex(
-                    TransferRedisChannels.transfer_result_key(event.transfer_id),
-                    300,
-                    payload,
-                )
+                # Best-effort, and isolated on purpose: persistence only covers
+                # the waiter that has not subscribed yet. A waiter that IS
+                # subscribed depends on the publish below, so a failure to write
+                # the durable copy must never skip it — that would turn a
+                # storage hiccup into a transfer timeout on a healthy pub/sub.
+                try:
+                    await redis.setex(
+                        TransferRedisChannels.transfer_result_key(event.transfer_id),
+                        300,
+                        payload,
+                    )
+                except Exception as e:
+                    logger.error(
+                        f"Failed to persist terminal transfer event for "
+                        f"{event.transfer_id}: {e}"
+                    )
             await redis.publish(channel, payload)
             logger.info(f"Published {event.type} event for {event.transfer_id}")
         except Exception as e:
--- a/api/services/telephony/providers/__init__.py
+++ b/api/services/telephony/providers/__init__.py
@@ -14,4 +14,5 @@ from api.services.telephony.providers import (  # noqa: F401  -- import for side
     twilio,
     vobiz,
     vonage,
+    voxpro,
 )
--- a/api/services/telephony/providers/voxpro/__init__.py
+++ b/api/services/telephony/providers/voxpro/__init__.py
@@ -1,0 +1,88 @@
+"""VoxPro (India VNO) telephony provider package.
+
+Registering this package adds VoxPro to Dograh's telephony provider list.
+VoxPro is a licensed Indian VNO offering managed PSTN (own 1400/1600 range) via
+its connector API — Twilio-style simplicity, India-native, DPDP-compliant.
+"""
+
+from typing import Any, Dict
+
+from api.services.telephony.registry import (
+    ProviderSpec,
+    ProviderUIField,
+    ProviderUIMetadata,
+    register,
+)
+
+from .config import VoxProConfigurationRequest, VoxProConfigurationResponse
+from .provider import VoxProProvider
+from .transport import create_transport
+
+
+def _config_loader(value: Dict[str, Any]) -> Dict[str, Any]:
+    return {
+        "provider": "voxpro",
+        "api_key": value.get("api_key"),
+        "tenant_id": value.get("tenant_id"),
+        "api_base": value.get("api_base"),
+        "from_numbers": value.get("from_numbers", []),
+    }
+
+
+_UI_METADATA = ProviderUIMetadata(
+    display_name="VoxPro (India VNO)",
+    docs_url="https://docs.dograh.com/integrations/telephony/voxpro",
+    fields=[
+        ProviderUIField(
+            name="api_key",
+            label="VoxPro API Key",
+            type="password",
+            sensitive=True,
+            description="Per-tenant API key issued by VoxPro for the connector API",
+        ),
+        ProviderUIField(
+            name="tenant_id",
+            label="Tenant ID",
+            type="text",
+            description="VoxPro tenant / X-Tenant-ID (e.g., AI_Katha_1783948668)",
+        ),
+        ProviderUIField(
+            name="api_base",
+            label="Connector Base URL",
+            type="text",
+            required=False,
+            description="VoxPro AI Connector base URL (leave default unless self-hosted)",
+        ),
+        ProviderUIField(
+            name="from_numbers",
+            label="Phone Numbers (DIDs)",
+            type="string-array",
+            description="VoxPro DIDs in E.164 without + prefix",
+        ),
+    ],
+)
+
+
+SPEC = ProviderSpec(
+    name="voxpro",
+    provider_cls=VoxProProvider,
+    config_loader=_config_loader,
+    transport_factory=create_transport,
+    transport_sample_rate=8000,
+    config_request_cls=VoxProConfigurationRequest,
+    ui_metadata=_UI_METADATA,
+    config_response_cls=VoxProConfigurationResponse,
+    account_id_credential_field="tenant_id",
+)
+
+
+register(SPEC)
+
+
+__all__ = [
+    "SPEC",
+    "VoxProConfigurationRequest",
+    "VoxProConfigurationResponse",
+    "VoxProProvider",
+    "create_transport",
+]
--- a/api/services/telephony/providers/voxpro/config.py
+++ b/api/services/telephony/providers/voxpro/config.py
@@ -6,7 +6,7 @@ API with a per-tenant API key, and VoxPro originates/bridges the call on its own
 carrier and streams Plivo/Twilio-standard mu-law audio to Dograh.
 """
 
-from typing import List, Literal, Optional
+from typing import List, Literal
 
 from pydantic import BaseModel, Field
 
--- a/api/services/telephony/providers/voxpro/config.py
+++ b/api/services/telephony/providers/voxpro/config.py
@@ -1,0 +1,41 @@
+"""VoxPro telephony configuration schemas.
+
+VoxPro is a licensed Indian VNO. Unlike self-hosted ARI, the Dograh user does
+NOT expose their own Asterisk — they authenticate to VoxPro's managed connector
+API with a per-tenant API key, and VoxPro originates/bridges the call on its own
+carrier and streams Plivo/Twilio-standard mu-law audio to Dograh.
+"""
+
+from typing import List, Literal, Optional
+
+from pydantic import BaseModel, Field
+
+DEFAULT_API_BASE = "https://connector.voxprosolutions.com"
+
+
+class VoxProConfigurationRequest(BaseModel):
+    """Request schema for VoxPro configuration."""
+
+    provider: Literal["voxpro"] = Field(default="voxpro")
+    api_key: str = Field(..., description="VoxPro connector API key (per-tenant)")
+    tenant_id: str = Field(
+        ..., description="VoxPro tenant / X-Tenant-ID (e.g., AI_Katha_1783948668)"
+    )
+    api_base: str = Field(
+        default=DEFAULT_API_BASE,
+        description="Base URL of the VoxPro AI Connector",
+    )
+    from_numbers: List[str] = Field(
+        default_factory=list,
+        description="VoxPro DIDs available for outbound caller-ID (E.164, no + prefix)",
+    )
+
+
+class VoxProConfigurationResponse(BaseModel):
+    """Response schema for VoxPro configuration with masked sensitive fields."""
+
+    provider: Literal["voxpro"] = Field(default="voxpro")
+    api_key: str  # Masked
+    tenant_id: str
+    api_base: str = DEFAULT_API_BASE
+    from_numbers: List[str]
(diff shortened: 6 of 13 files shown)


AManjunath-Voxpro:
Thanks @cubic-dev-ai and @greptile-apps — all valid, fixed in the latest push. Summary:

**cubic**
- **P0 (transport, serializer `auto_hang_up`)** — fixed. Pass `auto_hang_up=False` to the Plivo serializer. VoxPro terminates the call through its own connector (Asterisk), so no Plivo `auth_id`/`auth_token` are needed; with `auto_hang_up=True` the serializer `__init__` would `ValueError`. Verified both ways.
- **P1 (`X-Tenant-ID` header)** — fixed. `_headers()` now sends `X-Tenant-ID` alongside `X-API-Key` so the connector scopes by tenant.
- **P2 (`ws_url` `None` guard)** — fixed. `initiate_call` now raises if `workflow_run_id`/`workflow_id`/`organization_id` are missing.
- **P0 (media WS route not mounted)** — fixed. Now streams over the **generic mounted route** `/api/v1/telephony/ws/{workflow_id}/{organization_id}/{workflow_run_id}` (ids come via kwargs, same as Telnyx), which dispatches to `handle_websocket`. No custom route needed.
- **P3 (no-op test)** — fixed. `test_get_available_phone_numbers` now awaits and asserts the returned list.

**greptile**
- **Media URL 404** — same generic-route fix as above.
- **Status callback unmounted** — set `WEBHOOK_ENDPOINT = None`. VoxPro is a call-control provider (like `ari`); lifecycle is driven by the media WS + connector, so there is no HTTP status-callback route to 404.
- **Transfer targeted the tracking UUID** — fixed. `conference_name` is `transfer-{original_call_sid}` (per `pipecat_engine_custom_tools`), so `transfer_call` decodes it to the real carrier call id and transfers that. VoxPro controls Asterisk, so it does a native transfer rather than the conference bridge cloud providers need. Added a test asserting `/v1/calls/CARRIER123/transfer` (not the transfer_id).

Provider unit tests: **9 passing**. Happy to keep iterating — and re #596, a clean `pip`-only checkout still cannot collect the suite (missing published pipecat modules), so I ran these against the provider package in isolation.

AManjunath-Voxpro:
Thanks both for the re-review and the LGTM 🙏

Addressed the follow-up you flagged on the `transfer-` prefix coupling: extracted it into a documented module constant `TRANSFER_CONFERENCE_PREFIX` in `voxpro/provider.py`, with a comment pointing at the source of truth (`workflow/pipecat_engine_custom_tools.py`), and the transfer test now builds `conference_name` from that constant so the contract is exercised.

I kept the constant on the provider side rather than a truly shared one, since a shared constant would mean editing `pipecat_engine_custom_tools.py` — and `providers/AGENTS.md` explicitly says a provider should not need to touch core files. If you would prefer the naming contract to live in a shared location, happy to follow your steer. Still 9/9.

AManjunath-Voxpro:
Good catch @cubic-dev-ai — valid P2, and thanks for tracing it through `status_processor.py`. Fixed in the latest push, following the same pattern Telnyx uses:

The chain: transfer tool reads `workflow_run.gathered_context["call_id"]` (`pipecat_engine_custom_tools.py:767`), which `status_processor.py:205` normally fills from status callbacks. VoxPro has `WEBHOOK_ENDPOINT=None`, so that never ran and `call_id` stayed unset → `conference_name="transfer-None"`.

Fix (both call directions):
- **Outbound:** `initiate_call` now returns `call_id` in `provider_metadata`. The dispatcher merges `provider_metadata` into `gathered_context` (`campaign_call_dispatcher.py:382`), so `gathered_context["call_id"]` is set at initiation — exactly what Telnyx does (`provider_metadata={"call_id": ...}`).
- **Inbound** (no dispatcher): `handle_websocket` now persists the connector `call_id` from the WS `start` event via `update_workflow_run(gathered_context={"call_id": ...})`. That method deep-merges gathered_context, so it is safe. This is the WS-start-event path you suggested.
- Test asserts `provider_metadata["call_id"]`.

Still 9/9.

AManjunath-Voxpro:
Thanks @greptile-apps — that P1 is exactly right, and thanks for tracing it through the shared flow. Fixed in `ffc86d8`, following the seam Telnyx uses for the same problem.

**The gap:** after `transfer_call` returns, `pipecat_engine_custom_tools` mutes the pipeline and blocks on `wait_for_transfer_completion(transfer_id)` for a `TransferEvent` on the Redis channel. A blind carrier transfer completes asynchronously in the connector, and VoxPro published nothing → the wait ran to `timeout` → `transfer_failed` while the caller sat on hold music.

**Fix (Telnyx's `webhook_url` pattern):**
- `transfer_call` now hands the connector a `result_url` → new route `POST /api/v1/telephony/voxpro/transfer-result/{transfer_id}`. I deliberately did **not** publish from inside `transfer_call`: the shared flow subscribes only *after* `transfer_call` returns, so an inline publish would race the subscriber and be lost. The callback (fired when Asterisk actually completes the transfer) is safely after subscription — same reason Telnyx/ARI publish from their event paths, not from `transfer_call`.
- New `providers/voxpro/routes.py` publishes `DESTINATION_ANSWERED` on success or `TRANSFER_FAILED` with a reason. For a **blind** transfer the caller is already bridged to the destination in Asterisk, so `DESTINATION_ANSWERED` just makes the shared flow end the Dograh pipeline leg (`end_call_with_reason(TRANSFER_CALL)`) — no conference join, which VoxPro doesn't need. Idempotent via `claim_transfer_step` so retried callbacks publish exactly once. Auto-mounts through the registry — no core edits.

**@cubic-dev-ai nit:** the broad `except` in `handle_websocket` now logs the full traceback via `logger.opt(exception=True)` (a persistent `call_id`-persist failure would break transfer recovery, so it should be diagnosable, not silent). Kept it broad on purpose — a transient DB hiccup must not drop a live call.

Tests: **14 passing** — added `build_transfer_event` success/failure/no-context mapping, the transfer-result route's publish + idempotency, and a `result_url` assertion on `transfer_call`.

AManjunath-Voxpro:
You've read it exactly right, and thanks for keeping the two issues distinct.

Yes — the media-socket capability-token gap is intentionally **not** fixed in this PR, and it should be tracked as a follow-on. The reasoning:

- The trust boundary lives in the shared `_handle_telephony_websocket` (that `TODO(security)` is upstream of this PR), and **every** WebSocket-media provider — Telnyx, Vobiz, VoxPro — dials back the same `/ws/{workflow_id}/{organization_id}/{workflow_run_id}` route. So it's a cross-provider core concern, not a VoxPro one.
- Closing it means changing connect-time auth for all providers, which is out of scope for a provider-addition PR — `providers/AGENTS.md` explicitly says a provider shouldn't need to touch core files, and I didn't want to silently alter the trust model for the existing providers.
- `ffc86d8` doesn't widen the boundary: a peer that can reach that socket can already start/drive the run, and the `call_id` I persist comes from the same `start` event the pipeline already consumes — it's a subset of the existing capability, not a new sink.

Filed it as **#598** with a suggested direction (per-run HMAC/capability token minted alongside `ws_url` at `initiate_call` / inbound start, verified in `_handle_telephony_websocket`), so it lands once for all providers. And as you noted, the transfer-result route is UUID-scoped and unaffected.

Happy to take that one on as a separate PR against the shared handler if you'd like — just wanted to keep this PR to the VoxPro provider surface.

AManjunath-Voxpro:
Quick follow-up @greptile-apps: rather than leave the media-socket capability-token work as just a tracked issue, I've put up the fix as a separate draft PR against `main` — **#599** (closes #598).

It's the shared-handler approach we discussed: a stateless HMAC `?token=` minted by a new `ws_auth.build_media_ws_url()` at all 9 mint sites (telnyx/twilio/vonage/cloudonix/plivo/vobiz + the two inbound routes), verified in `_handle_telephony_websocket`. Opt-in and backward-compatible — no secret set → byte-for-byte the legacy URL; secret set → tokens minted + logged; `TELEPHONY_WS_TOKEN_ENFORCE=1` → invalid/missing rejected (4401). It deliberately stops at closing forgeability and leaves the one-shot-redemption + `initialized→running` CAS (the deeper half of your `TODO`) for a follow-up, since that one needs a run-creation schema change.

Kept it out of this PR so #595 stays scoped to the VoxPro provider; VoxPro will adopt the helper on rebase once both land. Reviews welcome on #599.

AManjunath-Voxpro:
Thanks for the thorough read, and agreed on every point — the static-token nuance most of all.

It's inherent to the carrier-callback model: the token has to live in a URL the carrier dials back, so it's a **capability**, not a session secret — it proves *possession of the server secret*, not freshness. A token leaking via a CDR or carrier log is replayable against **that one run** until the run ends (bounded, but real). That's precisely why the one-shot-redemption + `initialized→running` CAS is the right next step, and why I scoped it out here — it needs the run-creation schema change and closes a distinct state-race too. Called out in the #599 body so it isn't a surprise for anyone auditing.

I've taken **#599 out of draft** for your formal pass. As you noted, the finding on this PR closes fully once VoxPro swaps its f-string for `build_media_ws_url` on rebase after #599 lands. Appreciate the careful review throughout.

AManjunath-Voxpro:
@a6kme when you get a moment, this one would love a human review — it is mergeable with all automated checks green (cubic + greptile feedback addressed in-branch).

Since opening it we have also validated the provider end-to-end against our live Asterisk carrier stack: originate → media WS (Plivo-format frames) → blind transfer → transfer-result callback, verified on a real PSTN call through our Vodafone trunk. And following your pointer in #596, the full `tests/telephony/` suite passes on the host-managed setup (105 passed, submodule pipecat).

Happy to rebase onto latest main first if you would prefer.

a6kme:
Hello @AManjunath-Voxpro - Yes. This is on our agenda. A rebase of main would be helpful actually. Thanks a lot for your contribution.

AManjunath-Voxpro:
Rebased onto latest `main` — linear history, no merge commit. Thanks @a6kme!

Three things went in with the rebase:

**1. `generate_validation_error_response` was missing.** The shared `/inbound/run` dispatcher calls it on every provider class; VoxPro only defined `generate_error_response`, so a misconfigured inbound DID raised `AttributeError` inside the dispatcher and the caller got a bare `<Hangup/>` with the real reason hidden. Found by running a live inbound call against our Asterisk trunk — a number with no `inbound_workflow_id` assigned. Added with a regression test.

**2. `validate_phone_number`** (new abstract method on `main`) implemented against the connector's `/v1/numbers`. That list is the tenant's ownership record — an API key can only originate on its own trunk/DIDs — so it is the right authority rather than opting out. Non-PSTN addresses stay with the dialplan, as with the other PBX-managed providers.

**3. Media WS URL now goes through `ws_auth.build_media_ws_url`,** so the capability token from #599 is minted for VoxPro too (the guard test on `main` correctly caught this).

Plus the greptile P1 below. `tests/telephony/` is 160 passed on the rebased branch.

Also worth mentioning: the provider is now validated end to end against a live India carrier — outbound, blind transfer with the result callback, and inbound (DID → connector → `/inbound/run` → media WS → workflow run) all on real PSTN calls.

AManjunath-Voxpro:
> **P1 Transfer completion can be lost before the waiter subscribes**

Confirmed and fixed — good catch, and the ordering probe matches what we see on real calls: our connector reports the transfer outcome in roughly half a second, well before `transfer_call()` returns and the workflow reaches `wait_for_transfer_completion()`. The event went into an empty channel and the transfer timed out despite the destination answering, with the callback claim making a retry a no-op.

Fixed in the shared transfer manager rather than the VoxPro route, since the hazard belongs to the wait itself and every callback-driven provider shares it:

- `publish_transfer_event` persists terminal events (`DESTINATION_ANSWERED` / `TRANSFER_FAILED`) under `transfer:result:<transfer_id>` **before** publishing.
- `wait_for_transfer_completion` subscribes **first**, then reads that key. An event published before the subscription is read from the key; one published after arrives on the subscription; neither path can miss it.
- The key is cleared by `remove_transfer_context` and otherwise expires on the same 300s TTL as the context.

Regression tests in `tests/telephony/test_call_transfer_manager.py` reproduce your scenario directly: the pub/sub double never delivers, so the test can only pass via the persisted copy. Suite is 160 passed.

AManjunath-Voxpro:
> **P1 Persistence failure drops live event**

Right, and that one was mine — fixed in `c4888b5`.

The durable copy only rescues a waiter that has not subscribed yet; a waiter that is already subscribed depends entirely on the publish. Leaving the `setex` inside the shared `try` meant a Redis write error skipped publishing altogether and timed out a transfer whose pub/sub was healthy and whose callback had arrived — strictly worse than before the durable copy existed. Trading availability for durability, in the one direction that was already working.

Persistence is now isolated and best-effort: it logs and falls through, and `publish` always runs. Regression test drives a Redis double whose `setex` raises and asserts the event is still published on the transfer channel.

`tests/telephony/` is 161 passed.

AManjunath-Voxpro:
> **P1 security — Unauthenticated transfer callback completes calls**

Valid, and fixed in `175a156`. The docstring leaned on `transfer_id` being an unguessable capability (the reasoning Telnyx's transfer webhook_url uses), but that ID is logged and passed around, the blast radius is a transfer accepted as connected with no carrier confirmation, and VoxPro already has a signing scheme for its inbound webhook — so there was no good reason not to use it here.

Callbacks now carry `X-VoxPro-Signature`, HMAC-SHA256 of the raw body keyed on the tenant API key. Two details worth calling out:

- **The key is resolved through the transfer, not the request.** `TransferContext.workflow_run_id` → run → telephony configuration → credentials, so a caller cannot nominate which key verifies its own message. That is exactly what the `workflow_run_id` comment on TransferContext anticipated.
- **Verification runs before `claim_transfer_step`.** Otherwise an unauthenticated POST would burn the one-shot claim and lock out the genuine callback — a denial-of-completion even when the signature check rejects it. A transfer whose context has expired is now rejected rather than published: it cannot be authenticated and has nothing left to unblock.

Tests cover unsigned, wrong-key and unknown-transfer callbacks, and assert the claim is untouched on rejection. The connector signs the callback with the same scheme (verified on a live PSTN transfer: signature present and valid). `tests/telephony/` is 164 passed.

AManjunath-Voxpro:
Heads-up on the red `drift-check` here — it looks unrelated to this PR, so flagging it in case it is news.

It fails in the **Dump OpenAPI spec** step, before anything of ours is exercised:

```
ImportError: Blocked import of regex from current working directory for security reasons.
Use '-P' or set PYTHONSAFEPATH to prevent Python from searching the current working directory.
```

`nltk` (pulled in via `pipecat.utils.string`) refuses to import `regex` while the repo root is on `sys.path`. The same job fails the same way on `arnofrxdd/url-path-params-clean`, so it is environmental rather than branch-specific — recent runner/toolchain behaviour, by the look of it.

If it helps, the error names its own fix: running that step with `python -P` (or `PYTHONSAFEPATH=1` in the step env) keeps the CWD off `sys.path` and the import guard is satisfied. Happy to open a separate PR for the workflow file if you would like — I have left it alone here since it is outside this change.

The formatting drift this job caught earlier **was** ours and is fixed (`scripts/format.sh` applied; it also cleaned some pre-existing drift in `voxpro/config.py` and the voxpro tests). `pytest` is green.

AManjunath-Voxpro:
> **VoxPro callback behavior remains unverified** — the harness ended during application-router imports

Fair, and worth closing properly rather than leaving it on source inspection. Here is executed evidence for each of the four behaviours, run twice: as unit tests, and as live HTTP against a running Dograh (real Redis, real Postgres, this branch at `175a156`).

**Live app** — route mounted, from the running instance's OpenAPI:
```
["/api/v1/telephony/voxpro/transfer-result/{transfer_id}"]
```

Seeded a real `TransferContext` (workflow_run_id 2, resolving to a real telephony configuration) and posted callbacks to the live route:

```
1. unsigned                    -> HTTP 401
2. signed with a foreign key   -> HTTP 401
   claim key after both        -> EXISTS 0      (one-shot claim NOT burned by rejected callers)
3. correctly signed            -> HTTP 200 {"status":"ok","event":"destination_answered"}
4. durable copy after publish  -> EXISTS 1
   transfer:result:live-proof-001 =
   {"type": "destination_answered", "transfer_id": "live-proof-001",
    "original_call_sid": "CALLER1", "transfer_call_sid": "DEST1", ...}
5. same signed callback again  -> HTTP 200 {"status":"duplicate"}
```

And the case the durability fix exists for — a waiter that subscribes **after** the callback already landed (pub/sub would have dropped it):

```
LATE WAITER RESULT: destination_answered
  transfer_id      : live-proof-001
  transfer_call_sid: DEST1
```

**Unit tests** covering the same paths, all green:
```
TestTerminalEventDurability::test_completion_published_before_wait_is_still_delivered PASSED
TestTerminalEventDurability::test_terminal_event_is_persisted_before_publish          PASSED
TestTerminalEventDurability::test_removing_context_clears_stored_result               PASSED
TestPersistenceFailureDoesNotDropLiveEvent::test_publish_still_happens_when_persist_fails PASSED
test_transfer_result_rejects_unsigned_callback     PASSED
test_transfer_result_rejects_wrong_key_signature   PASSED
test_transfer_result_rejects_unknown_transfer      PASSED
test_transfer_result_route_is_idempotent           PASSED
test_transfer_result_route_publishes_event         PASSED
```

Full suite: `tests/telephony/` 164 passed. Probe keys were removed from Redis afterwards.

Separately, the transfer path is exercised on real PSTN calls through our carrier — the connector signs the callback and the last live run logged `CALLBACK SIGNATURE: present=True valid=True` before the completion event.

chewwbaka:
Thanks for contributing the VoxPro integration, @AManjunath-Voxpro 

Before reviewing implementation correctness or completeness, we need the following information to decide whether VoxPro should become an officially supported Dograh provider:

1. Explain how users create a VoxPro account, complete KYC, obtain a phone number, and connect it to Dograh.
2. Provide a Dograh-controlled sandbox tenant and securely share credentials so we can independently test inbound calls, outbound calls, hang-up, and transfers.
3. Provide reproducible smoke-test evidence for successful inbound, outbound, and transfer flows.
4. Clarify who will maintain the integration and what support and escalation channels will be available to Dograh and its users.
5. Document the connector API used by this PR, since it differs from VoxPro’s publicly documented JWT-based API.
6. Add the missing Mintlify setup page referenced by the integration.
7. Separate or explain the shared transfer-manager changes and Pipecat submodule update, as they extend beyond the core provider integration.

Without these clarifications, we cannot decide whether to accept VoxPro as a supported integration or proceed with the technical review.

chewwbaka:
Closing this because of the blocking requirements.

PR #599, opened 2026-07-30
feat(telephony): capability-token auth for the media WebSocket (#598)

### LINKED ISSUE #598: Telephony media WebSocket authenticates peers only by the {workflow_id}/{organization_id}/{workflow_run_id} triple (shared across all providers)
### Summary

The telephony media WebSocket authenticates connecting peers **solely** by possession of the numeric id triple in the URL:

```
/api/v1/telephony/ws/{workflow_id}/{organization_id}/{workflow_run_id}
```

`_handle_telephony_websocket` (`api/routes/telephony.py`) has no signature/capability check, so the triple is a **guessable bearer capability**. An untrusted peer who supplies (or guesses) a valid triple can open the socket, start the run, and drive `gathered_context` (e.g. inject a `call_id` via the provider's `start` event).

The handler already acknowledges this:

```python
# TODO(security): ``organization_id`` arrives in the URL the provider dials
# back, so it is caller-supplied and unauthenticated — this socket has no
# signature check, and the id triple is a guessable bearer capability.
```

### Why a separate issue

This is **shared core**, not provider-specific: Telnyx, Vobiz, VoxPro and every other WebSocket-media provider dial back the same route, so the trust boundary lives in `_handle_telephony_websocket`, not in any one provider. It predates PR #595 (VoxPro provider) — surfaced during that review by @greptile-apps — and fixing it means changing behavior for **all** providers, which is out of scope for a single provider-addition PR (`providers/AGENTS.md`: a provider shouldn't need to edit core files).

### Suggested direction

Mint a per-run capability token when the backend hands the provider its `ws_url` (at `initiate_call` / inbound `start_inbound_stream`) and verify it on connect — e.g. an HMAC over `{workflow_id}:{organization_id}:{workflow_run_id}` with a server secret, appended as a `?token=` query param, checked in `_handle_telephony_websocket` before the run starts. This keeps the id triple in the path (providers already build it) while closing the unauthenticated-peer gap for every provider at once.

Scoped-and-fine counterexample from the same review: the VoxPro transfer-result callback (`/voxpro/transfer-result/{transfer_id}`) is keyed by an unguessable UUID, so it isn't affected — this issue is specifically about the id-triple media socket.

Refs: #595

AManjunath-Voxpro:
## What & why

Closes the media-WebSocket trust-boundary gap tracked in #598 (surfaced during the VoxPro provider review, #595). The socket

```
/api/v1/telephony/ws/{workflow_id}/{organization_id}/{workflow_run_id}
```

is dialed back by the carrier/connector, so the id triple in its URL is a **guessable bearer capability** — `_handle_telephony_websocket`'s own `TODO(security)` calls this out. It affects **every** WebSocket-media provider, so the fix lives in the shared handler + a shared URL builder, not in any one provider.

## Approach — stateless HMAC capability token, opt-in, backward-compatible

- **`ws_auth.build_media_ws_url()`** mints the URL with `?token=HMAC-SHA256(secret, "{wid}:{oid}:{rid}")` when `TELEPHONY_WS_TOKEN_SECRET` is set. **With no secret it returns the exact legacy URL**, so switching a provider onto the helper is a no-op until an operator opts in.
- All **9 mint sites** — telnyx, twilio, vonage, cloudonix, plivo, vobiz, and the two inbound routes in `routes/telephony.py` — now build the URL through the helper.
- **`_handle_telephony_websocket`** verifies the token, with a two-phase zero-downtime rollout:

| `TELEPHONY_WS_TOKEN_SECRET` | `TELEPHONY_WS_TOKEN_ENFORCE` | Behavior |
|---|---|---|
| unset (default) | — | feature off, URLs unchanged |
| set | off (default) | tokens minted + verified; invalid ones **logged**, connection allowed |
| set | on | invalid/missing token **rejected** (WS close `4401`) |

So an operator sets the secret first (everything gets a valid token, spoofs get logged), confirms the logs are clean, then flips enforcement.

## Scope / non-goals

This closes **forgeability** — a peer can no longer connect by guessing ids, only by holding the server secret. It intentionally does **not** implement the one-shot-redemption + `initialized -> running` compare-and-swap the `TODO(security)` sketches: that needs a run-creation schema change and also closes a separate state-race, so it's better as its own PR. Happy to follow up with it if you'd like it in the same effort.

Note: **VoxPro (#595)** builds the same URL and will adopt `build_media_ws_url` on rebase once both land — kept out of this diff so this PR stands alone against `main`.

## Tests

`api/tests/telephony/test_ws_auth.py` — 5 cases: mint/verify roundtrip, tamper across each id, secret rotation invalidation, no-secret no-op URL, enforcement flag. The handler gate itself is a small verified branch in the diff (an integration test would want your DB/websocket fixtures).

Refs #598

---
## Summary by cubic
Add HMAC capability-token auth to the telephony media WebSocket and cover the ARI path to prevent guessable URL access. Opt-in and backward-compatible with staged enforcement. Closes #598.

- **New Features**
  - Added `ws_auth.build_media_ws_url` that appends `?token=HMAC-SHA256(secret, "{workflow_id}:{organization_id}:{workflow_run_id}")`; with no `TELEPHONY_WS_TOKEN_SECRET`, it returns the legacy URL.
  - `_handle_telephony_websocket` verifies tokens; logs invalid when `TELEPHONY_WS_TOKEN_ENFORCE` is off, rejects with 4401 when on; non-ASCII tokens are treated as invalid (no 500).
  - All mint sites now use the helper (telnyx, twilio, vonage, cloudonix, plivo, vobiz, and inbound routes); ARI external-media now mints/forwards the token in `v()` params so `/ws/ari` authenticates under enforcement.
  - Tests cover mint/verify, tamper cases, secret rotation, no-secret no-op URL, enforcement flag, non-ASCII token invalidation, and ARI str/int id equivalence.

- **Migration**
  - Set `TELEPHONY_WS_TOKEN_SECRET` to start minting and verifying tokens (connections still allowed).
  - Monitor logs for unverifiable connections.
  - Enable `TELEPHONY_WS_TOKEN_ENFORCE=true` to reject invalid or missing tokens.

<sup>Written for commit deeb1e20aee27a6d09b5ac2fbc8b7b333e235bcd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/599?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

Adds optional HMAC capability-token authentication for telephony media WebSockets. Carrier provider URLs, inbound media URLs, and ARI external-media callbacks now include a token when configured, while installations without a token secret retain the existing URLs.

The focused execution checks exercised token generation and identifier binding, ARI string/integer identifier compatibility, enforced rejection of missing, malformed, non-ASCII, and mismatched tokens, and all changed URL emitters. The predecessor's non-ASCII token path raised a `TypeError`; the current implementation safely rejects that input and closes an enforced connection as unauthorized. No defects were found in the exercised behavior.

Merge safety: safe to merge.

**What T-Rex did**
- The review-authored WebSocket token contract harness was run against the current repository, exiting with code 0 and exercising token minting, verification, provider URL construction, no-secret compatibility, the shared WebSocket handler under valid, missing, malformed, non-ASCII, and identifier-mismatched inputs, ARI external-media transport data, and all changed emitters, while also revalidating the predecessor verifier at commit 6874d68 where a non-ASCII token raised TypeError and confirming the current code rejects that token with a 4401 Unauthorized response.
- A before-and-after assessment shows the prior non-ASCII failure path at commit 6874d68 is fixed, and the harness run demonstrates all capability-token contracts pass with exit code 0 after the PR.
- Artifacts created to support the review include the standalone PR 599 WebSocket token contract validation script and the baseline and post-PR validation outputs captured by the harness runs.

<a href="https://app.greptile.com/trex/runs/16596644/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(telephony): ARI token minting + non-..."](https://github.com/dograh-hq/dograh/commit/deeb1e20aee27a6d09b5ac2fbc8b7b333e235bcd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48706302)</sub>

### DIFF
--- a/api/constants.py
+++ b/api/constants.py
@@ -99,6 +99,20 @@ POSTHOG_HOST = os.getenv("POSTHOG_HOST", "https://us.i.posthog.com")
 ENABLE_ARI_STASIS = os.getenv("ENABLE_ARI_STASIS", "false").lower() == "true"
 SERIALIZE_LOG_OUTPUT = os.getenv("SERIALIZE_LOG_OUTPUT", "false").lower() == "true"
 
+# ── Telephony media WebSocket authentication ────────────────────────────────
+# The carrier/connector dials back the media socket
+# /api/v1/telephony/ws/{workflow_id}/{organization_id}/{workflow_run_id}, whose
+# id triple is otherwise a guessable bearer capability. When a secret is set,
+# providers append an HMAC ``?token=`` to that URL and the handler verifies it
+# (see api/services/telephony/ws_auth.py). Two-phase, backward-compatible rollout:
+#   * secret unset            -> feature off, URLs unchanged (default)
+#   * secret set, enforce off -> tokens minted + verified; invalid ones only logged
+#   * secret set, enforce on  -> invalid/missing tokens rejected (WS close 4401)
+TELEPHONY_WS_TOKEN_SECRET = os.getenv("TELEPHONY_WS_TOKEN_SECRET") or None
+TELEPHONY_WS_TOKEN_ENFORCE = (
+    os.getenv("TELEPHONY_WS_TOKEN_ENFORCE", "false").lower() == "true"
+)
+
 # Logging configuration
 LOG_FILE_PATH = os.getenv("LOG_FILE_PATH", None)
 LOG_LEVEL = os.getenv("LOG_LEVEL", "DEBUG").upper()
--- a/api/routes/telephony.py
+++ b/api/routes/telephony.py
@@ -38,6 +38,7 @@ from api.services.telephony.factory import (
     get_telephony_provider_by_id,
     get_telephony_provider_for_run,
 )
+from api.services.telephony import ws_auth
 from api.services.telephony.transfer_event_protocol import (
     TransferEvent,
     TransferEventType,
@@ -609,6 +610,28 @@ async def _handle_telephony_websocket(
         # Set the run context
         set_current_run_id(workflow_run_id)
 
+        # Capability-token check. The id triple in the URL is otherwise a
+        # guessable bearer capability (see the TODO above and ws_auth.py). This
+        # is a no-op until an operator sets TELEPHONY_WS_TOKEN_SECRET; once set,
+        # invalid tokens are logged, and rejected only when enforcement is on.
+        if ws_auth.token_configured():
+            token = websocket.query_params.get("token")
+            if not ws_auth.verify_ws_token(
+                workflow_id, organization_id, workflow_run_id, token
+            ):
+                if ws_auth.enforcement_enabled():
+                    logger.warning(
+                        f"[telephony ws] rejecting unauthenticated connection for "
+                        f"run {workflow_run_id} (org {organization_id})"
+                    )
+                    await websocket.close(code=4401, reason="Unauthorized")
+                    return
+                logger.warning(
+                    f"[telephony ws] UNVERIFIED media socket for run {workflow_run_id} "
+                    f"(org {organization_id}); allowed because TELEPHONY_WS_TOKEN_ENFORCE "
+                    f"is off — set it to enforce"
+                )
+
         # Get workflow run to determine provider type
         workflow_run = await db_client.get_workflow_run(
             workflow_run_id, organization_id=organization_id
@@ -881,9 +904,8 @@ async def handle_inbound_run(request: Request):
                 )
 
             backend_endpoint, wss_backend_endpoint = await get_backend_endpoints()
-            websocket_url = (
-                f"{wss_backend_endpoint}/api/v1/telephony/ws/"
-                f"{workflow_id}/{config.organization_id}/{workflow_run_id}"
+            websocket_url = ws_auth.build_media_ws_url(
+                wss_backend_endpoint, workflow_id, config.organization_id, workflow_run_id
             )
 
             return await provider_instance.start_inbound_stream(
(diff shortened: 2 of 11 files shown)


AManjunath-Voxpro:
Both valid — fixed in `deeb1e2`. Thanks, the ARI one especially is a sharp catch.

**P1 (ARI enforcement outage) — confirmed and fixed.** `/ws/ari` routes through the same `_handle_telephony_websocket`, and the ARI manager builds its routing params via `v(workflow_id,organization_id,workflow_run_id)` with no token — so `TELEPHONY_WS_TOKEN_ENFORCE` would have 4401'd every ARI media call before the provider could start. Rather than exempt ARI, I mint the **same** token into the `v()` params in `ari_manager._create_external_media`, so Asterisk echoes it back as `?token=` and the shared check passes. Enforcement now covers all media paths uniformly. The str-vs-int id types line up because `f"{1}" == f"{'1'}"` (the ARI path mints with string ids, the handler verifies with the int triple) — added a regression test pinning that equivalence.

**P2 (non-ASCII token) — confirmed and fixed.** `hmac.compare_digest` raises `TypeError` on non-ASCII `str` args, and `token` is attacker-controlled, so a crafted value would 500 the socket (1011) and skip the reject/audit path. Now compared as bytes with `TypeError`/`UnicodeError` treated as an invalid token — malformed tokens take the normal path. Added a test with non-ASCII input.

Tests now 7/7.

PR #667, opened 2026-08-15
Adds asterisk call ID to initial_context

jasonhamilton:
When working with the agents, for some reason the call_id would not populate in the gathered_context.  I believe this gets filled in later in the call.  I have a tool call that requires the call_id in order to interact with the ARI directly but was unable to do so.  I feel like this is a very important variable to have in the initial_context since without it there is no way to associate an inbound channel with an exiting call that was transferred into the system.

---
## Summary by cubic
Adds `call_id` to `initial_context` for inbound ARI calls so tools and agents can access it immediately. Previously `initial_context` omitted `call_id`; now both `initial_context` and `gathered_context` include it, enabling direct ARI interactions and proper association of inbound channels with transferred calls.

- Review: Change is in `api/services/telephony/ari_manager.py` within `_handle_inbound_stasis_start`, adding `call_id` to `initial_context`. Verify any consumers reading `initial_context` handle the new `call_id` field.
- Migration: If you enforce a strict schema for `initial_context`, add `call_id` to the allowed fields.

<sup>Written for commit 2a57a06c142edc9070fb2fd317bc1a7575757f9e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/667?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

Adds the inbound Asterisk channel identifier to a workflow run’s initial context while preserving the existing gathered-context value. A mocked inbound StasisStart execution compared the parent revision with this change: the updated code created one workflow run with both `initial_context.call_id` and `gathered_context.call_id` set to the stable handler channel ID, rather than the mismatching event value. No defects were found; the change is safe to merge.

The inbound ARI workflow-run creation path correctly exposes the stable channel identifier in both initial and gathered context.

A focused mocked execution exercised the changed inbound-call path, captured the workflow-run creation arguments, and confirmed the new context value matches the established gathered-context identifier.

**Files Needing Attention:** No files need follow-up attention.

**What T-Rex did**
- The focused ARI inbound call validation harness was executed against the parent revision and the PR revision to inspect how call IDs flow into workflow runs.
- Before the PR, the inbound ARI workflow-run creation showed gathered\_context.call\_id set to stable-ari-channel-42 and initial\_context.call\_id was null.
- After the PR, the inbound ARI workflow-run creation showed both gathered\_context.call\_id and initial\_context.call\_id equal to stable-ari-channel-42.
- This confirms that the initial context now uses the stable handler channel identifier.

<a href="https://app.greptile.com/trex/runs/19061708/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["Adds asterisk call ID to initial\_context"](https://github.com/dograh-hq/dograh/commit/2a57a06c142edc9070fb2fd317bc1a7575757f9e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53675859)</sub>

### DIFF
--- a/api/services/telephony/ari_manager.py
+++ b/api/services/telephony/ari_manager.py
@@ -887,6 +887,7 @@ class ARIConnection:
                     "provider": "ari",
                     "telephony_configuration_id": self.telephony_configuration_id,
                     "external_pbx_call": external_pbx_call,
+                    "call_id": call_id,
                 },
                 gathered_context={
                     "call_id": call_id,


chewwbaka:
HI @jasonhamilton 

> I have a tool call that requires the call_id in order to interact with the ARI directly

Can you explain your usecase? Specifically, where will the call id be consumed in the workflow (for example, a prompt, HTTP tool, or webhook), whether this applies to inbound, outbound, or transferred calls, and which Asterisk identifier the downstream system expects.

We need this context to confirm the correct behavior and scope before we can proceed with the PR.

a6kme:
@jasonhamilton - To add on top of @chewwbaka's comment, I believe we allow accessing the variables from gathered_context also in preset parameters of tool calls - so you can use gathered_context.call_id there. If its not accessible, thats a separate bug which I can look into. 

This is a convention that if we change for ARI, we would have to change for all the other providers to keep it consistent. 

So please let us know the use case and we would be happy to fix the issue.

jasonhamilton:
@chewwbaka I am using the Asterisk Rest Interface in order to transfer the call to an extension for example: http://asterisk_url/ari/channels/{{call_id}}/continue.  This allows transferring the call using the existing asterisk channel.  This allows the call to retain its original context.  The original goal was to retain the caller ID, however it is common to set global variables on a channel for other purposes.  The variable that is received looks like:  "call_id": "1787077259.1902", which is the channel the inbound caller is on.  

If I try and transfer the call back to asterisk using the ARI tool, there seems to be a couple of problems that I see. The first is that associating the call back with the original call/channel is back is quite a challenge since there is no context passed back the system.  The second issue with the ARI tool is that as far as I can tell it creates a new bridge and joins the channels instead of passing the call off so the asterisk system is coupled to Dograh even after the handoff.

@a6kme I was trying to use the gathered_context variable but the only data present was the nodes_visited array.  The funny thing is that after working on this for a bit and then restarting the API service, the gathered_context came back and I was able to access it.  I'm not sure if there is some sort of race condition or another issue but I tried many times to access the call info from gathered_context.  I even created additional nodes to try and give it some time to load it but still ran into problems.  If the call_id can be taken from the gathered context and that works consistently that's fine.

This may be a bit off topic but It would be great to be able to pass arbitrary context from Asterisk to Dograh.  For example, if asterisk already had customer data, or customer ID you can pass this on and use this data within Dograh.

chewwbaka:
@a6kme https://github.com/dograh-hq/dograh/pull/692

chewwbaka:
Addressed the changes in #692 which is merged; hence closing this now.

PR #668, opened 2026-08-16
fix: Enforce met node transition conditions

### LINKED ISSUE #641: [Bug]: Agent not moving to next node even if conditions are met. [Extraction variables + prompt not enforcing move]
### Describe the bug

In the multi Node agent, the agent is not moving to the next note, even if the conditions defined at a particular note has been met. Extraction variables and prompt both are not able to enforce the agent to move to the next node.

### Steps to Reproduce

Create a agent with multiple nodes, enable extraction variables and node conditions, call the agent, meet condition and you would see the model is not moving through the next node.

### Expected behavior

Model will not move to the next node all the time. 

### Screenshots

_No response_

### Device details

macOS

### Additional context

_No response_

mvanhorn:
Multi-node agents can remain on the current node after an outgoing edge condition has become true, even when the node prompt also directs the model to move on. The runtime currently exposes each edge condition as a transition tool description, but the composed system prompt contains no explicit rule requiring the model to call a matching transition tool. This makes a documented workflow invariant depend on the model inferring mandatory routing behavior from tool metadata alone. Extraction variables do not provide a routing signal here: project documentation says they are collected after a node completes and are not available in node prompts.

## Summary by cubic

Extend node system-prompt composition with a concise, provider-neutral transition instruction whenever the active node has outgoing edges. The instruction should tell the model to evaluate the live conversation against the available transition conditions and immediately call the matching transition tool once a condition is met, without hardcoding any issue-specific condition or coupling routing to post-node extraction. Keep the existing global prompt, node prompt, and recording-mode instruction ordering intact, and do not add transition guidance to nodes with no outgoing edges.

## T-Rex validation blocked

- Compose a prompt for a node with one outgoing edge and verify the node prompt plus mandatory transition guidance are present.
- Compose a prompt for a node with multiple outgoing edges and verify the guidance remains generic while the edge-specific conditions continue to live in the generated tool schemas.
- Compose a prompt for an end or otherwise edge-less node and verify no transition guidance is appended.
- Compose a prompt with global and recording-mode content enabled and verify the new guidance does not replace or reorder those existing instructions.

Fixes #641

---
## Summary by cubic
Adds a mandatory transition instruction to node system prompts so agents immediately call a matching transition tool when an outgoing edge condition is met. Previously, prompts did not require a transition call, so agents could stay on the current node after a condition became true.

- Appends `NODE_TRANSITION_INSTRUCTIONS` only when the node has outgoing edges; end or edge-less nodes remain unchanged.
- Preserves prompt ordering: global prompt → node prompt → transition instruction → recording-mode instructions.
- Leaves edge-specific conditions in transition tool schemas; no routing via extraction variables or post-node flow.
- Tests cover single/multiple edges, absence of edges, and prompt ordering to validate behavior.

Fixes Linear issue #641.

<sup>Written for commit a534949c5b87c001a26558177f407f0fbc1d4772. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/668?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

This change adds mandatory transition guidance to workflow-node prompts when outgoing transitions are available. It preserves the existing global prompt, node prompt, and recording-mode instruction order, while keeping transition conditions in the corresponding tool schemas.

A focused runtime harness confirmed that multi-edge nodes receive ordered transition guidance and retain both condition-bearing transition tools. Nodes without outgoing edges receive neither transition guidance nor transition tools. No defects were found, and the change is safe to merge.

### T-Rex validation blocked
The repository’s focused pytest target could not complete collection because the `dotenv` package is missing from the environment. The standalone composition harness executed successfully against the changed implementation.

The verified prompt and tool-composition behavior matches the intended workflow transition behavior and is safe to merge.

No independent defects remain. An executed harness loaded the changed composer and exercised prompt ordering, multiple transition conditions, and edge-less nodes; the unavailable pytest collection does not contradict those observed results.

**Files Needing Attention:** No files need follow-up changes. The focused pytest environment needs the missing `dotenv` package before the repository test can be run there.

**What T-Rex did**
- Ran the transition-composition harness for the PR's Pipecat engine context composer and confirmed the runtime prompt order global → node → transition → recording, with guidance and tools appearing only where transitions exist.
- Verified the exact harness run reported global=0, node=23, transition=76, recording=296, and TRANSITION\_GUIDANCE\_PRESENT=True, confirming the ordered prompt sequence and the presence of two condition-bearing transition tools; this shows the PR preserves the runtime schemas.
- The focused pytest run was blocked by a missing dotenv dependency, preventing test execution and explaining why the runtime path could not complete in this environment.

<a href="https://app.greptile.com/trex/runs/19082604/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: Enforce met node transition conditi..."](https://github.com/dograh-hq/dograh/commit/a534949c5b87c001a26558177f407f0fbc1d4772) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53719851)</sub>

### DIFF
--- a/api/services/workflow/pipecat_engine_context_composer.py
+++ b/api/services/workflow/pipecat_engine_context_composer.py
@@ -45,6 +45,10 @@ RULES:
 - Use ▸ when you need to generate a dynamic, contextual response.
 - *NEVER* mix modes in a single response, since we rely on the markers to decide whether to play using TTS or Pre-recorded audio."""
 
+NODE_TRANSITION_INSTRUCTIONS = """\
+NODE TRANSITION INSTRUCTIONS - MANDATORY:
+Evaluate the live conversation against the conditions described by the available transition tools. As soon as a condition is met, immediately call the matching transition tool."""
+
 
 def compose_system_prompt_for_node(
     *,
@@ -55,9 +59,9 @@ def compose_system_prompt_for_node(
 ) -> str:
     """Compose the full system prompt text for a workflow node.
 
-    Combines the global prompt, node-specific prompt, and (when recordings
-    are enabled anywhere in the workflow) the recording response mode
-    instructions into a single string.
+    Combines the global prompt, node-specific prompt, transition instructions,
+    and (when recordings are enabled anywhere in the workflow) the recording
+    response mode instructions into a single string.
 
     Args:
         node: The workflow node to compose the prompt for.
@@ -76,6 +80,9 @@ def compose_system_prompt_for_node(
     formatted_node_prompt = format_prompt(node.prompt)
 
     parts = [p for p in (global_prompt, formatted_node_prompt) if p]
+
+    if node.out_edges:
+        parts.append(NODE_TRANSITION_INSTRUCTIONS)
 
     if has_recordings and "RECORDING_ID:" in formatted_node_prompt:
         parts.append(RECORDING_RESPONSE_MODE_INSTRUCTIONS)
--- a/api/tests/test_pipecat_engine_context_composer.py
+++ b/api/tests/test_pipecat_engine_context_composer.py
@@ -1,0 +1,126 @@
+from types import SimpleNamespace
+
+import pytest
+
+from api.services.workflow.pipecat_engine_context_composer import (
+    RECORDING_RESPONSE_MODE_INSTRUCTIONS,
+    compose_functions_for_node,
+    compose_system_prompt_for_node,
+)
+
+
+TRANSITION_GUIDANCE = """\
+NODE TRANSITION INSTRUCTIONS - MANDATORY:
+Evaluate the live conversation against the conditions described by the available transition tools. As soon as a condition is met, immediately call the matching transition tool."""
+
+
+def make_edge(name: str, condition: str) -> SimpleNamespace:
+    return SimpleNamespace(
+        condition=condition,
+        get_function_name=lambda: name,
+    )
+
+
+def make_node(
+    prompt: str,
+    *,
+    out_edges: list[SimpleNamespace],
+    add_global_prompt: bool = False,
+) -> SimpleNamespace:
+    return SimpleNamespace(
+        prompt=prompt,
+        out_edges=out_edges,
+        add_global_prompt=add_global_prompt,
+        document_uuids=None,
+        tool_uuids=None,
+    )
+
+
+def make_workflow(global_prompt: str | None = None) -> SimpleNamespace:
+    global_node_id = "global" if global_prompt else None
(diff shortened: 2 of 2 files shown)


chewwbaka:
Hi @mvanhorn, thanks for your PR. 
I understand the rationale here that edge conditions are currently exposed in transition-tool descriptions, but the composed prompt does not explicitly require a matching tool call; and extraction variables cannot provide a live routing signal because they are collected after node completion.

Before we add this mandatory instruction to every node with outgoing edges, have we verified that the reported failure cannot be addressed at the **node-authoring level**?

In practice and our experience, builders are more likely to describe a transition’s full completion criteria in the originating node prompt than in the transition condition. Could you share a minimal reproduction showing that the originating node already:

1. defines the complete criteria for when it is finished; and
2. explicitly tells the model to call the exact transition tool exposed at runtime once those criteria are satisfied?

For example:

Success criteria:
- If the user denies they are {{initial_context.first_name}}, call 'end_call_apology'.
- Only if the user confirms they are {{initial_context.first_name}} and explicitly consents to the survey, call 'collect_personal_details'.
- If the user explicitly refuses to talk, call 'end_call'.

This seems safer than globally instructing the model to transition "as soon as a condition is met." The proposed instruction makes the transition-tool condition the effective routing authority, but those conditions may be short or ambiguous while the node prompt deliberately includes further collection, confirmation, or exception handling.

If this authoring pattern fixes the reported case, the appropriate follow-up may be documentation, templates, and builder guidance rather than a global system-prompt change. If it does not, please share the minimal workflow/prompt and transcript demonstrating why explicit node-level completion criteria plus an explicit transition-tool call still fail.

chewwbaka:
Closing the PR as we await evidence/ steps to reproduce for the linked issue. Once we have that we can work on the approach. So far,  a prompt authoring pattern with documentation, templates, and builder guidance seems a better alternative to a global system-prompt change. Happy to reconsider this again later if needed.

PR #740, opened 2026-09-07
fix(kb): migrate vector index from ivfflat to hnsw (#635)

### LINKED ISSUE #635: [Bug]: KB vector search recall silently collapses on small corpora — ivfflat index built on empty table + default probes=1
## Environment

- Self-hosted OSS deployment, dograh **v1.44.0** (docker compose, `pgvector/pgvector:pg17`)
- Knowledge base: 9 documents / ~340 chunks, ingested through the normal upload → process flow (`retrieval_mode: "chunked"`, embeddings `openai` / `text-embedding-3-small`)

## Summary

Knowledge-base vector search silently returns the wrong chunks on small corpora. Two compounding causes, both in how the `ix_kb_chunks_embedding_ivfflat` index is set up:

1. The ivfflat index is created **by the migration, on an empty table**, with a hardcoded `lists = 100`. IVF centroids are computed at index build time, so they are meaningless for whatever data is ingested later — and nothing ever rebuilds the index.
2. The application never sets `ivfflat.probes`, so every query runs with pgvector's default **`probes = 1`**. With ~340 vectors spread over 100 lists (~3–4 vectors per list), each query effectively considers only a handful of candidates.

The failure is completely silent: ingestion reports `completed`, embeddings are valid, and searches return plausible-looking results — just not the relevant ones.

## Reproduction

1. Fresh deployment, ingest ~9 documents (~340 chunks). Ours were Japanese shelter listings; any corpus of this size should do.
2. One document contains a chunk whose contextualized text literally matches the query topic (a facility list for the 出水 district).
3. `POST /api/v1/knowledge-base/search` with `{"query": "中央区 出水 周辺の避難場所", "limit": 3, "document_uuids": [<all 9 uuids>]}`.

**Before fix** — the matching chunks never appear; top results are unrelated chunks from other documents (top similarity 0.51), and some query phrasings return **0 results** despite `limit=3` and 340 candidate rows:

```
sim=0.511  [general-knowledge doc]  (unrelated section)
sim=0.473  [general-knowledge doc]  (unrelated section)
sim=0.352  [preparedness doc]       (unrelated section)
```

**After fix** (`REINDEX INDEX ix_kb_chunks_embedding_ivfflat; ALTER DATABASE postgres SET ivfflat.probes = 100;`) — the same query immediately returns the correct chunks:

```
sim=0.688  [ward doc]  出水中学校 …
sim=0.684  [ward doc]  出水南中学校 …
sim=0.678  [ward doc]  出水小学校 …
```

Verified with `SELECT` that the "missing" chunks existed all along with valid 1536-dim embeddings and the correct `embedding_model` value — only the ANN index stood between them and the results.

## Impact

- Every self-hosted install with a small-to-medium knowledge base (hundreds to low thousands of chunks — arguably the typical case) gets severely degraded retrieval out of the box.
- The degradation presents as "the RAG just isn't very good", so users are unlikely to diagnose it as an index configuration problem.
- Recall gets *worse* after bulk re-imports (centroids go stale), which punishes exactly the users who refresh their KB content.

## Suggested fixes

Any of these would resolve it:

1. **Switch the index to HNSW** (`USING hnsw (embedding vector_cosine_ops)`, pgvector ≥ 0.5): no training step, so no stale-centroid problem, and much better recall on small corpora. This is the current community default for this use case.
2. Set `ivfflat.probes` per query/session based on corpus size (e.g. `SET LOCAL ivfflat.probes = ...` in `search_similar_chunks`).
3. At minimum: size `lists` dynamically and re-create the index after ingestion, and document the `probes` caveat for self-hosters.

dasocreative:
Fixes #635

### Summary
Migrates the vector similarity index on `knowledge_base_chunks.embedding` from `ivfflat` to `hnsw` (`vector_cosine_ops`).

Building `ivfflat` on an empty table with default `probes = 1` causes vector search recall to collapse on small-to-medium corpora. `hnsw` avoids the empty-table centroid problem and provides reliable recall out of the box.

### Changes
- **Model**: Updated `KnowledgeBaseChunkModel` in `api/db/models.py` to use `ix_kb_chunks_embedding_hnsw`.
- **Migration**: Added reversible Alembic migration to swap `ivfflat` for `hnsw`.
- **Tests**: Added unit tests in `api/tests/test_kb_hnsw_index.py` verifying the index definition and migration metadata.

### Testing
- Ran `pytest api/tests/test_kb_hnsw_index.py` (passed).
- Verified format and OpenAPI drift with `./scripts/format.sh`.

## Checklist
- [x] I have read and followed the [contributing guidelines](https://github.com/dograh-hq/dograh/blob/main/CONTRIBUTING.md).

---
## Summary by cubic
Migrates the vector index on `knowledge_base_chunks.embedding` from `ivfflat` to `hnsw`. The old `ivfflat` index was built on an empty table, which collapses search recall on small-to-medium corpora with pgvector's default `probes = 1`. `hnsw` builds an incremental graph without pre-clustering, so it avoids the empty-table centroid problem and delivers reliable recall out of the box.

**Changes**
- `api/db/models.py` declares `ix_kb_chunks_embedding_hnsw` with `postgresql_using="hnsw"` and `vector_cosine_ops`.
- The reversible Alembic migration drops the legacy index and creates the HNSW one; downgrade restores `ivfflat` with `lists=100`.
- Unit tests in `api/tests/test_kb_hnsw_index.py` verify the model index options and migration revision chain.

<sup>Written for commit d4b6f8f3ec6aa460945ecaca2b7da7390e17730a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h3>Greptile Summary</h3>

## Summary

This change replaces the knowledge-base IVFFlat vector index with HNSW and adds index-definition tests.

Two deployment-blocking issues need resolution before merging:
- Existing databases using pgvector 0.4.x cannot create the new HNSW index, and the migration removes the existing IVFFlat index before failing.
- Building the replacement index on a populated table can stop knowledge-base writes for the duration of the build and may exceed the pre-upgrade job deadline.

## T-Rex validation blocked

A live lock-duration reproduction was unavailable because the environment lacked Python 3.13, Alembic dependencies, PostgreSQL client/server tools, and a usable database service.

<h3>Confidence Score: 3/5</h3>

Not safe to merge until the pgvector compatibility path and production-safe index-build strategy are addressed.

The pgvector compatibility failure was reproduced against pgvector 0.4.4. The write-blocking issue is established by the non-concurrent index operation, Alembic transaction configuration, and deployment deadline, although a live lock-duration reproduction could not run in this environment.

**Files Needing Attention:** api/alembic/versions/a1c9e8f4b3d7_migrate_kb_chunks_vector_index_to_hnsw.py needs a pgvector-version preflight or upgrade path and a maintenance-aware index creation strategy. The related Alembic transaction configuration and migration-job deadline must also be considered.

**What T-Rex did**
- T-Rex produced a proof for a posted P1 finding and linked it to the corresponding review comment.
- T-Rex performed an environment check for a live migration lock-duration reproduction and confirmed that Python 3.13, Alembic dependencies, PostgreSQL tools, and a usable database service were unavailable.
- T-Rex produced a second proof for a posted P1 finding.
- T-Rex validated HNSW compatibility by comparing before/after states, inspecting migration history, and confirming that HNSW was introduced in pgvector 0.5.0.
- T-Rex executed the HNSW index locking evidence script and observed a static check pass, with live repro blocked by the environment constraints.

<a href="https://app.greptile.com/trex/runs/22965088/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=4"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=4">View all artifacts</picture></a>

</details>

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. General comment 

   <a href="#">P1</a> **HNSW migration blocks databases that still run pgvector 0.4.x**

   - **Bug**
     - `upgrade()` drops the existing IVFFlat index and unconditionally attempts to create an HNSW index. A real pgvector 0.4.4 database rejects that DDL with `ERROR: access method "hnsw" does not exist`, causing the Alembic upgrade to stop after the IVFFlat index is dropped.
   - **Cause**
     - HNSW was introduced by pgvector 0.5.0, but the migration assumes it is already available. It has neither a version preflight nor an extension-upgrade step, and no deployment compatibility policy enforces pgvector >=0.5 for pre-existing databases.
   - **Fix**
     - Before dropping the IVFFlat index, explicitly require `vector` extension version >=0.5 and fail with an actionable upgrade instruction, or provide/document a safe operator-managed `ALTER EXTENSION vector UPDATE` path before this migration. Do not silently proceed on older versions; preserve the existing index if the prerequisite is absent.

2. General comment 

   <a href="#">P1</a> **Build the HNSW index outside the migration transaction without blocking writers**

   - **Bug**
     - `upgrade()` drops the old index and then executes a plain `op.create_index()` for `knowledge_base_chunks.embedding`. On an existing populated table, PostgreSQL's non-concurrent `CREATE INDEX` takes a `SHARE` lock, which blocks concurrent `INSERT`, `UPDATE`, and `DELETE` until the HNSW build completes. The migration is run inside Alembic's configured transaction, so switching this call directly to `CREATE INDEX CONCURRENTLY` would fail because PostgreSQL prohibits that command in a transaction block. The Helm migration job executes pre-upgrade and is capped at 600 seconds, so a large build also risks aborting the release deadline.
   - **Cause**
     - The index operation at lines 34-41 has neither `postgresql_concurrently=True` nor explicit autocommit SQL, while `api/alembic/env.py:90-93` places each revision in `context.begin_transaction()`.
   - **Fix**
     - Split the online index build from transactional DDL: create the HNSW index with `CREATE INDEX CONCURRENTLY` through Alembic's autocommit block (or an explicitly non-transactional migration path), then drop the old index concurrently in a separate safe step. Account for the interval in which both indexes exist, and set an operationally appropriate migration deadline or schedule the build during a controlled maintenance window.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(kb): migrate vector index from ivffl..."](https://github.com/dograh-hq/dograh/commit/d4b6f8f3ec6aa460945ecaca2b7da7390e17730a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61334349)</sub>

> Greptile also left **2 inline comments** on this PR.

**Context used:**

- Knowledge Base — [HTTP APIs and persistence](https://app.greptile.com/dograh-hq/-/custom-context/knowledge-base/dograh-hq/dograh/-/docs/backend-api-and-persistence.md)
- Knowledge Base — [AI configuration and knowledge processing](https://app.greptile.com/dograh-hq/-/custom-context/knowledge-base/dograh-hq/dograh/-/docs/ai-and-knowledge.md)


dasocreative:
Thanks for the review! Here’s why we actually need HNSW and why it fits in the current codebase:
The existing IVFFlat index was created on an empty table, locking in bad centroids and hurting search recall as new chunks arrive. HNSW indexes dynamically as data is added. It has been standard in pgvector since v0.5.0 and the current codebase runs 0.7+.
Why not CONCURRENTLY: Postgres forbids CONCURRENTLY inside transaction blocks. Bypassing Alembic’s transaction breaks rollback safety if the migration fails. Keeping migrations atomic is standard across Dograh.
Let me know if you need an ALTER EXTENSION vector UPDATE; pre-check if you'd like to support legacy local setups!

PR #760, opened 2026-09-12
fix: wait for idle farewell playback before hanging up

### LINKED ISSUE #750: [Bug]: Realtime LLMs — user-idle disconnect hangs up before the farewell is spoken
### Describe the bug

When a workflow uses a **speech-to-speech model** (OpenAI Realtime, Grok Realtime; the same path applies
to Gemini Live / Azure Realtime) and the caller stays silent through both user-idle timeouts, Dograh is
supposed to say a short goodbye and then hang up. In practice the caller hears **silence, or at most a
clipped syllable, and then the line drops.**

The idle handler asks the model for the farewell and, in the very next statement, ends the call. With a
cascaded LLM + TTS pipeline that works, because the farewell text and audio are generated *inside* the
pipeline ahead of the `EndFrame`. With a realtime model the farewell is produced **asynchronously over
the provider websocket**, while the `EndFrame` travels through the pipeline in milliseconds and the
realtime service's `stop()` closes that websocket immediately. The response is cancelled before any
audio arrives.

The max-call-duration path is unaffected by this report: it deliberately uses `CancelFrame`
(`abort_immediately=True`) and never promises a farewell.

### Steps to Reproduce

1. Create a workflow with a realtime model (e.g. OpenAI Realtime `gpt-realtime-2`) and a normal greeting node.
2. Set `max_user_idle_timeout` to something short (e.g. 10 s) to speed the test up.
3. Start a call (WebRTC test call or any telephony provider) and **say nothing** after the greeting.
4. After the first timeout the agent asks "Are you still there?" (this part works). Keep silent.
5. After the second timeout the call ends with disposition `user_idle_max_duration_exceeded`.
6. Listen: no farewell is played (or only a fragment). In the API log, directly after
   `Handling user_idle, attempt: 2`, note `Finishing run with reason: user_idle_max_duration_exceeded`
   within ~1 ms and the realtime service's `TTFB: 0.00xs` metric — the TTFB is stopped by the disconnect
   before any audio arrived.
7. Repeat with a cascaded workflow (any STT + LLM + TTS): the farewell plays in full before the hang-up.

### Expected behavior

The farewell requested by the second idle prompt ("wish them a good day") is spoken to completion, and
only then is the call ended with `user_idle_max_duration_exceeded`. Realtime and cascaded pipelines should
behave the same here.

### Screenshots

_No response_

### Device details

_No response_

### Root cause (verified against code)

1. **The handler ends the call synchronously right after requesting the farewell.**
   `api/services/workflow/pipecat_engine_callbacks.py:42-62` — on `_retry_count >= 2` it pushes an
   `LLMMessagesAppendFrame([...wish them a good day...], run_llm=True)` and immediately awaits
   `engine.end_call_with_reason(EndTaskReason.USER_IDLE_MAX_DURATION_EXCEEDED.value)`.

2. **`end_call_with_reason` queues an `EndFrame` at once.**
   `api/services/workflow/pipecat_engine.py:735-784` — sets `_mute_pipeline = True`, awaits pending
   variable extraction, then `self.task.queue_frame(EndFrame(reason=...))` (a `CancelFrame` only when
   `abort_immediately=True`). Nothing waits for the bot to finish speaking.

3. **Realtime services disconnect on `EndFrame` without draining the in-flight response.**
   `pipecat/src/pipecat/services/openai/realtime/llm.py:503-510` — `stop(EndFrame)` calls
   `super().stop()` then `_disconnect()`; `_disconnect()` (`llm.py:746-754`) sets `_disconnecting`,
   calls `stop_all_metrics()` and closes the websocket. The Dograh wrapper's
   `_handle_messages_append` (`api/services/pipecat/realtime/openai_realtime.py:106-131`) has only just
   sent `conversation.item.create` + `response.create`; the model's audio would arrive 0.5-1 s later, and
   by then the socket is gone. The Grok wrapper follows the same pattern.

4. **Why cascaded pipelines are fine.** There the `LLMMessagesAppendFrame` becomes an `LLMContextFrame`
   that the LLM processes *in pipeline order*; the LLM text, TTS audio and the `EndFrame` all flow
   through the same queues, so the `EndFrame` cannot overtake the farewell. The realtime case breaks that
   ordering because the farewell comes back on a side channel.

### Evidence (local dev logs, `logs/20260818_002551/`)

Three calls that reached the second idle expiry. In each, the TTFB metric is stopped by the
disconnect; **0.001-0.005 s means no audio had been received**. The transport tears down 2-4 s later.

```
uvicorn-worker-0.log  (Grok Realtime)
23:49:58.356  pipecat_engine_callbacks:handle_idle:45   Handling user_idle, attempt: 2
23:49:58.357  pipecat_engine:end_call_with_reason:780  Finishing run with reason: user_idle_max_duration_exceeded, ... queueing frame EndFrame
23:49:58.357  pipeline.worker:_wait_for_pipeline_end   PipelineWorker#6: Closing. Waiting for EndFrame#1(reason: user_idle_max_duration_exceeded)
23:49:58.358  frame_processor_metrics:stop_ttfb_metrics DograhGrokRealtimeLLMService#3 TTFB: 0.001s
23:50:00.591  smallwebrtc.transport:disconnect          Disconnecting to Small WebRTC

uvicorn-worker-17.log  (OpenAI Realtime)
23:06:46.854  Handling user_idle, attempt: 2
23:06:47.651  Finishing run with reason: user_idle_max_duration_exceeded ...
23:06:47.653  DograhOpenAIRealtimeLLMService#2 TTFB: 0.772s      <- first audio arrived ~20 ms before the socket closed
23:06:51.880  Disconnecting to Small WebRTC

uvicorn-worker-12.log  (OpenAI Realtime)
03:17:25.780  Handling user_idle, attempt: 2
03:17:25.783  Finishing run with reason: user_idle_max_duration_exceeded ...
03:17:25.790  DograhOpenAIRealtimeLLMService#28 TTFB: 0.005s
03:17:30.090  Disconnecting to Small WebRTC
```

The 0.8 s gap between attempt 2 and "Finishing run" in the second call is the awaited variable
extraction inside `end_call_with_reason`; it is the only reason a fragment of audio got through there.

### Suggested fix

Defer the `EndFrame` until the farewell has actually been spoken, with a bounded fallback so a
non-responding model cannot keep the call open. Dograh-side, no pipecat change needed:

```python
# pipecat_engine_callbacks.py — second expiry
await aggregator.push_frame(LLMMessagesAppendFrame([message], run_llm=True))
await self._engine.end_call_after_bot_turn(
    EndTaskReason.USER_IDLE_MAX_DURATION_EXCEEDED.value, fallback_secs=8.0
)

# pipecat_engine.py
async def end_call_after_bot_turn(self, reason: str, *, fallback_secs: float):
    self._pending_end_reason = reason
    self._pending_end_task = self.task.create_task(self._end_call_fallback(fallback_secs))

async def _end_call_fallback(self, secs: float):
    await asyncio.sleep(secs)
    await self.end_call_with_reason(self._pending_end_reason)

# in the existing frame hook that already tracks bot speech (should_mute_user sees
# BotStartedSpeakingFrame / BotStoppedSpeakingFrame):
elif isinstance(frame, BotStoppedSpeakingFrame):
    ...
    if self._pending_end_reason and self._bot_has_spoken_since_pending_end:
        cancel fallback task; await self.end_call_with_reason(self._pending_end_reason)
```

Guard details: only fire on a `BotStoppedSpeakingFrame` that follows a `BotStartedSpeakingFrame` seen
*after* the request (so a stale stop from the "are you still there?" turn cannot end the call early),
keep `_mute_pipeline` semantics unchanged, and let the fallback timer cover the case where the model
never answers. (Sketch, not tested.)

Alternative in pipecat: make the realtime services' `stop(EndFrame)` wait for `response.done` plus
output-queue drain with a timeout before `_disconnect()`. More invasive and provider-by-provider, so
the engine-level deferral is the smaller change.

### Related

The same race applies to any realtime response that is immediately followed by an `EndFrame`
(e.g. end-call tools that speak and then end). Not covered by this report, but the same deferral
mechanism would fix it.

### Confidence

**High.** The ordering is visible in code (steps 1-3 above) and the three logs show the disconnect
landing 1-800 ms after the farewell was requested, before or at the first audio byte.

mvanhorn:
Reuse the engine's existing `arm_speech_playback` and `wait_for_speech_playback` primitives, whose events are fed by `should_mute_user`, instead of introducing another provider-specific response tracker. Add a narrowly scoped engine operation such as `defer_end_call_until_bot_playback(reason, fallback_secs=8.0)` that synchronously arms playback and owns a single background waiter before the idle callback awaits pushing its farewell `LLMMessagesAppendFrame`. The waiter must let transport frame processing continue, wait for a fresh start followed by a stop, and invoke the existing `end_call_with_reason` only on completion or expiry of an eight-second total playback deadline.

On the second user-idle timeout, Dograh asks the model to say goodbye and immediately calls `end_call_with_reason`, which queues an `EndFrame` after final extraction. Realtime providers return audio asynchronously, so that termination can close the provider connection before the farewell reaches the caller. The reporter supplies reproduction steps and three log excerpts showing missing or clipped audio; cascaded LLM/TTS workflows normally preserve playback through pipeline ordering. The requested change covers the idle farewell for both pipeline types, while the hard max-duration callback retains its existing `abort_immediately=True` cancellation behavior.

Extend `api/tests/test_user_idle_handler.py` with focused async tests using a real minimal `PipecatEngine`, a fake frame sink, and mocked final extraction. Keep synchronization event-driven; shorten the production timeout via the operation's parameter or a module constant in tests, rather than waiting eight seconds.
First idle expiry still sends the existing realtime-compatible append frame with `run_llm=True` and does not arm termination. Active conversation resets and answer-supervisor suppression retain their current behavior.
On second expiry, prompt delivery returns before simulated realtime audio arrives. Assert no `EndFrame` or final extraction before playback, no termination on `BotStartedSpeakingFrame`, and exactly one graceful terminal frame after `BotStoppedSpeakingFrame`, with `user_idle_max_duration_exceeded` as the call status and unchanged disposition fallback rules.
Arm during an earlier utterance, then deliver only its stale stop: the call remains open. Deliver a fresh start/stop and observe completion. Also emit the new start/stop inside the fake prompt push before it returns, proving arming occurs early enough.
No audio, a start without a stop, and failed/no-output prompt delivery all reach bounded fallback without leaked tasks. Verify the playback deadline does not cancel the subsequent final extraction or terminal-frame queueing.
Repeated idle callbacks while waiting produce one farewell and one waiter. A stop arriving at the timeout boundary produces one disposal. User hangup, pipeline error, or the max-duration callback during the wait cancels the waiter, retains the winning reason and frame type, and does not add the farewell timeout to that termination path.
Cleanup cancels and awaits the pending task without queuing an idle end. The waiter's own successful ending does not self-cancel before its terminal frame is queued.
Add a pipeline-level regression in this same test file using existing mock transport/processor patterns: a fake realtime service returns immediately from the append request and produces audio later on a separate task. Observe playback completion before pipeline teardown. Retain a cascaded mock LLM/TTS case showing the farewell plays completely and the normal idle flow still ends.
Downstream validation: with Python 3.13, the pipecat submodule/dependencies, PostgreSQL/Redis test services, and `api/.env.test` configured, run `python -m pytest api/tests/test_user_idle_handler.py api/tests/test_transfer_message_playback.py api/tests/test_realtime_message_append.py api/tests/test_user_muting_during_bot_speech.py api/tests/test_answer_supervisor_wiring.py api/tests/test_termination_funnel.py`. Source only the test environment for pytest. A live workflow smoke test can confirm the issue's silent-caller sequence on OpenAI/Grok realtime and a cascaded workflow; report actual evidence rather than treating mocks as a live provider test.

## Checklist

- [x] I have read and followed the [contributing guidelines](https://github.com/dograh-hq/dograh/blob/main/CONTRIBUTING.md).

Fixes #750

---
## Summary by cubic
Fixes the silent-caller issue where the idle farewell could be cut off on realtime providers because the call ended as soon as the farewell prompt was pushed. The engine now waits for the farewell audio to complete (or an 8-second fallback deadline) before hanging up.

- Reuses the existing `arm_speech_playback` and `wait_for_speech_playback` primitives; adds `defer_end_call_until_bot_playback` which arms playback and owns a single background waiter.
- The idle handler now uses this operation instead of directly calling `end_call_with_reason`; other termination paths (user hangup, max duration) still take precedence and cancel the waiter.
- If the waiter's termination raises, it logs the error, clears `_call_disposed` and `_shutdown_task`, and retries with `PIPELINE_ERROR`/abort so a terminal frame is still queued.
- Adds focused tests for realtime deferred audio, duplicate idle callbacks, fallback deadlines, and cascaded workflows.

<sup>Written for commit a863facc51e467baf28a4e2bae943ee2dd86fa2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/760?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h2><a href="https://app.greptile.com/api/retrigger?id=63415672"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge.

<h3>Summary</h3>

This change defers idle-farewell termination until bot playback completes or reaches its timeout, while preserving a reliable immediate-abort fallback when graceful termination fails.

<sub>Reviews (4) · Last reviewed commit: ["fix: keep the idle farewell retry workin..."](https://github.com/dograh-hq/dograh/commit/a863facc51e467baf28a4e2bae943ee2dd86fa2d)</sub>


mvanhorn:
Good catch, that one is real. `_end_call_after_bot_playback` runs as an unawaited task, and `end_call_with_reason` flips `_call_disposed` before extraction and `queue_frame`, so a non-TimeoutError out of it killed the task silently with the user already muted and no terminal frame queued. A later termination attempt then no-ops on `_call_disposed`.

Fixed by catching `Exception` around the `end_call_with_reason` call, logging it with `exc_info=True`, clearing `_call_disposed`, and re-invoking with `EndTaskReason.PIPELINE_ERROR` and `abort_immediately=True`. `PIPELINE_ERROR` is already on the list of statuses that skip extraction, so the recovery path cannot re-trigger the extract that failed in the first place. `CancelledError` is a `BaseException` on 3.13, so genuine cancellation still propagates rather than being turned into a pipeline error.

Added `test_idle_farewell_queues_terminal_frame_if_end_call_raises`, which makes extraction raise and asserts a `CancelFrame(reason=pipeline_error)` is still queued and the call ends disposed. It fails on the old code with the bare `RuntimeError`.

Also rebased on main to clear the conflict; it was an additive collision in `pipecat_engine.py` where main added `queue_text_message` at the same point this branch added `defer_end_call_until_bot_playback`, so both are kept.

`pytest api/tests/test_user_idle_handler.py` is 21 passed.

PR #797, opened 2026-09-21
fix: set an Outbound Voice Profile on auto-created Telnyx Call Control Applications

### LINKED ISSUE #778: [Bug]: Auto-created Telnyx Call Control Application has no Outbound Voice Profile, so every outbound call fails with D38
## Summary

When a Telnyx telephony configuration is saved without a `connection_id`, Dograh auto-creates a Call Control Application in `_ensure_connection_id` (`api/services/telephony/providers/telnyx/__init__.py`, lines 53-60). The POST body only sets `application_name` and `webhook_event_url`:

```python
body = {
    "application_name": application_name,
    "webhook_event_url": inbound_url,
}
```

Telnyx Call Control Applications require an Outbound Voice Profile to place outbound calls. A freshly auto-created app has `outbound.outbound_voice_profile_id: null`, so the first outbound dial via `POST /v2/calls` fails with Telnyx error D38.

## Reproduction

1. Create a Telnyx telephony configuration without `connection_id` (Dograh auto-creates the app).
2. Add a phone number and set it as default caller ID (`ready_for_outbound` shows `true`).
3. `POST /api/v1/telephony/initiate-call`.

Actual result (live, verified 2026-09-16):

```json
{"data": {"call_control_id": "v3:..."},
 "errors": [{"code": "10010", "detail": "Connection has no Outbound Profile assigned D38. Connection has no Outbound Profile assigned."}],
 "telnyx_error": {"error_code": "D38"}}
```

The UI reports the config as ready for outbound even though the call can never succeed, because the setup-checklist hook only checks for a caller ID number, not the app's outbound profile.

## Suggested fix

Set an Outbound Voice Profile on the auto-created app. Options:

1. In `_ensure_connection_id`, include `"outbound": {"outbound_voice_profile_id": "<profile_id>"}` in the POST body, picking the org's first Outbound Voice Profile (or creating a dedicated one) via `GET /v2/outbound_voice_profiles`.
2. Surface the missing profile in the provider's `setup_checklist` so the UI shows "outbound blocked" before a dial fails.
3. Optionally retry `initiate_call` once with the profile attached when Telnyx returns D38, as a self-healing path for configs saved before this fix.

Manual workaround that unblocked the live test:

```
PATCH /v2/call_control_applications/{id}
{"application_name": "...", "outbound": {"outbound_voice_profile_id": "<existing-profile-id>"}}
```

Note the field is nested under `outbound`, not a top-level `outbound_profile_id` on the application resource.

## Environment

- dograh 1.47.0 (d91e4f5a), OSS local deployment
- Telnyx live API, 2026-09-16

a692570:
## Summary

Fixes #778.

When a Telnyx telephony configuration is saved without a `connection_id`, `_ensure_connection_id` auto-creates a Call Control Application. Until now the app was created with only `application_name` and `webhook_event_url`, so `outbound.outbound_voice_profile_id` stayed null and the very first outbound dial failed with Telnyx error D38 ("Connection has no Outbound Profile assigned"), even though the setup checklist reported the config as ready for outbound.

Verified live on 2026-09-16 against the Telnyx API: `POST /v2/calls` returned HTTP 403 with `telnyx_error.error_code: "D38"` for the auto-created app; attaching a profile via the nested `outbound` object and retrying succeeded end to end.

## What this does

- New helper `_fetch_first_outbound_voice_profile_id` lists the account's Outbound Voice Profiles with the same api key and returns the first id, or `None`.
- `_ensure_connection_id` now includes `"outbound": {"outbound_voice_profile_id": <id>}` in the app-create POST body when a profile exists.
- When the account has no profile, the app is still created and saved (inbound is unaffected) and a warning points at Mission Control. Profiles are never auto-created here, they are a billing-affecting resource.
- Profile-list failures (network or non-200) are logged and swallowed so the config save never breaks because of the enrichment.

## Tests

- New unit tests assert on the captured app-create POST body: includes the nested `outbound` key when a profile exists, omits it when the list is empty, and omits it when the list call raises.
- `pytest tests/telephony/telnyx/ -q` green (18 passed on the branch).

## Checklist

- [x] I have read and followed the contributing guidelines.

---
## Summary by cubic
Fixes auto-created Telnyx Call Control Applications so they now attach the account's first Outbound Voice Profile, preventing the first outbound call from failing with Telnyx error D38. If the account has no profile, the application is still created and a warning is logged; profile listing failures and unexpected payload shapes are swallowed so the config save never breaks. No profiles are ever auto-created, as they affect billing.

<sup>Written for commit dd44245e7b46d7d6cb48ef63898025d8652768c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/797?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h2><a href="https://app.greptile.com/api/retrigger?id=66827487"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

Safe to merge.

<h3>Summary</h3>

This change enriches automatically created Telnyx Call Control Applications with an existing outbound voice profile when available, while preserving inbound-only setup when no profile can be used.

<sub>Reviews (2) · Last reviewed commit: ["fix: set outbound voice profile on auto-..."](https://github.com/dograh-hq/dograh/commit/dd44245e7b46d7d6cb48ef63898025d8652768c0)</sub>


a692570:
Addressed the Greptile finding in a follow-up commit (dd44245e):

- `_fetch_first_outbound_voice_profile_id` now catches `asyncio.TimeoutError` alongside `aiohttp.ClientError` (verified: bare `asyncio.TimeoutError` from the default client timeout is not a `ClientError` subclass in aiohttp 3.13.3, so it escaped the old except and aborted the config save, contradicting the helper's documented "save must still succeed" contract).
- The decoded payload is now shape-validated (dict payload, list `data`, dict entries); any unexpected shape is treated as "no profiles available" with a warning instead of raising.

New tests cover the timeout path and the unexpected-payload shapes. Suite: 20 passed in `tests/telephony/telnyx/`.

a6kme:
Pushed `2a33af64a` (rebased onto main past the merged #799, which had conflicted in `test_provider.py`).

**One real bug: the "never breaks the save" contract had a hole.**

`_fetch_first_outbound_voice_profile_id` caught `(aiohttp.ClientError, asyncio.TimeoutError)`, but `await response.json()` on a 200 carrying `Content-Type: application/json` with a truncated or empty body raises `json.JSONDecodeError` — a `ValueError`, not an aiohttp error. It escaped both that handler and `_ensure_connection_id`'s `except aiohttp.ClientError`, so it would 500 the user's config save and persist no Call Control Application. Ironic given the careful defensive ladder below it for payload *shape*; the one failure that actually escaped wasn't covered, and no test exercised a malformed body. Now caught, with a test.

**Disabled profiles were still being bound.** `profiles[0]` took the first entry unconditionally. Telnyx profiles carry `enabled`, and binding a disabled one still fails the dial — so it's no better than binding nothing, while looking like success. The helper (now `_select_outbound_voice_profile`) skips disabled entries and takes the first usable one.

**The chosen profile wasn't logged.** The PR reasons carefully that profiles are billing-affecting and so refuses to *create* one — but then silently *picked* one, and only the `else` branch logged anything. Since accounts commonly have several (rate decks, regions, per-customer), a later rejection for a destination the bound profile doesn't whitelist was undiagnosable. It now logs the id, name, and `whitelisted_destinations` at bind time.

**Minor:** added a 10s timeout on the profile lookup. The existing test asserted `asyncio.TimeoutError` was tolerated, but nothing ever set a timeout — `aiohttp.ClientSession()` defaults to 300s total, so a hung Telnyx API could have stalled a config save for five minutes. Also requests `page[size]=100` so a usable profile isn't missed behind a short first page. And ran `./scripts/format.sh` (this was the red `drift-check`).

`pytest api/tests/telephony/ -q` green (382 passed).

**Worth a separate ticket, not fixed here:** this only fixes the auto-create path. `_ensure_connection_id` returns early when `connection_id` is already set, so every application auto-created before this PR — plus every user who pasted their own app id — still hits D38. And since Telnyx registers no `setup_checklist_resolver`, `get_setup_checklist` falls through to `caller_id_only_checklist`, which reports `ready_for_outbound` green regardless. The exact symptom quoted in the PR description is unchanged for existing users; a checklist step probing the application's `outbound.outbound_voice_profile_id` would fix the class rather than one entry point.

PR #798, opened 2026-09-21
fix: return a uniform 404 from the Telnyx events route so unsigned callers cannot enumerate run ids

### LINKED ISSUE #780: [Bug]: Telnyx events route returns 404 for unknown run IDs before signature verification, enabling run-ID enumeration
## Summary

In `handle_telnyx_events` (`api/services/telephony/providers/telnyx/routes.py`), the workflow-run lookup happens BEFORE webhook signature verification:

```python
# lines ~80-103 (current order)
workflow_run = await db_client.get_workflow_run_by_id(workflow_run_id)
if not workflow_run:
    raise HTTPException(status_code=404, detail="Workflow run not found")
...
provider = await get_telephony_provider_for_run(...)
signature_valid = await provider.verify_inbound_signature(...)
if not signature_valid:
    raise HTTPException(status_code=401, detail="Invalid webhook signature")
```

The signature check (merged in #271) correctly runs before any status mutation, but because the existence check comes first, an unauthenticated caller can distinguish "run ID exists" (401 Invalid webhook signature) from "run ID does not exist" (404 Workflow run not found) without any valid signature.

## Reproduction (live, 2026-09-16)

```
$ curl -X POST https://<host>/api/v1/telephony/telnyx/events/999999
-> 404 {"detail": "Workflow run not found"}

$ curl -X POST https://<host>/api/v1/telephony/telnyx/events/524   # existing run
-> 401 {"detail": "Invalid webhook signature"}
```

Different responses for signed vs unsigned endpoints let an external party probe valid workflow-run IDs. Run IDs are sequential integers, so the enumeration is trivial.

## Expected result

For an unauthenticated/unsigned request, the endpoint should return the same response regardless of whether the run ID exists, so run existence is not observable. Suggested options:

1. Verify the signature before the run lookup when the request carries no valid signature, returning 401 uniformly; or
2. Return a uniform 404 for both unknown runs and failed signature verification (as long as no state mutation happens on the 404 path, which is already the case).

The existing test `test_telnyx_events_route_verifies_signature_before_status_update` covers the mutation-ordering invariant and should keep passing under either option.

## Environment

- dograh 1.47.0 (d91e4f5a)
- Severity: low (info disclosure on run IDs, no data exposure, no state mutation)

a692570:
## Summary

Fixes #780.

In `handle_telnyx_events`, the run-existence 404 fires before signature verification, so an unsigned caller can distinguish existing runs (401 "Invalid webhook signature") from nonexistent ones (404 "Workflow run not found"). Run ids are sequential integers, which makes the enumeration trivial.

## What this does

- The invalid-signature response is now byte-identical to the unknown-run response: `404 "Workflow run not found"`.
- The existing `logger.warning` still names the real reason, so operators debugging a misconfigured `webhook_public_key` can see it in server logs. The route docstring documents the deliberate trade-off.
- No state mutation happens on either path (unchanged from #271's invariant).

Known limitation, left out of scope on purpose: #786 (in flight) returns 400 for provider-resolution failures after the lookups, so an org without a default telephony configuration still reveals run existence through the 404-vs-400 distinction. Full uniformity would need to absorb #786.

Verified live on 2026-09-16 (before this change): `POST /telephony/telnyx/events/999999` → 404, `POST /telephony/telnyx/events/524` (existing run) → 401. After the change, both return the identical 404.

## Tests

- New `test_events_route_uniform_404.py`: existing run + bad signature → 404; unknown run → 404; the two responses are identical.
- The existing `test_telnyx_events_route_verifies_signature_before_status_update` keeps passing.
- Full api suite green on this branch (3239 passed; one pre-existing timing flake in `test_campaign_outbound_capacity.py` passes standalone on both branch and main).

## Checklist

- [x] I have read and followed the contributing guidelines.

---
## Summary by cubic
Closes the run-existence side channel on the Telnyx events route: invalid-signature requests now return the same 404 as unknown runs, and an optional capability token makes the reject path constant-time so unsigned callers can't distinguish runs by response time either.

- The invalid-signature 404 is byte-identical to the unknown-run 404; the real reason stays in server logs.
- When `TELEPHONY_WS_TOKEN_SECRET` is set, the provider mints an HMAC `?token=` on the events webhook URL and the route verifies it before any database lookup, so existing and missing runs respond identically in body, status, and timing.
- With no secret configured, behavior falls back to the uniform-404 flow.
- New tests cover token minting/verification, the pre-auth gate, and 404 uniformity.

<sup>Written for commit 307bff1ff5c367db6a23ef2299ac0aebdd51eaa3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/dograh-hq/dograh/pull/798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

<h2><a href="https://app.greptile.com/api/retrigger?id=66827510"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 2/5</h2>

Unsafe to merge: valid lifecycle callbacks for active pre-deployment calls are rejected after token enforcement is enabled.

<h2>Findings</h2>

1. P1&nbsp;**Preserve In\-Flight Callbacks** <a href="https://github.com/dograh-hq/dograh/pull/798#discussion_r4066467748">▶</a>
2. P1&nbsp;Security&nbsp;**Timing Still Reveals Runs** <a href="https://github.com/dograh-hq/dograh/pull/798#discussion_r4059845916">▶</a>
3. P2&nbsp;**Test the HTTP Response** <a href="https://github.com/dograh-hq/dograh/pull/798#discussion_r4059845923">▶</a>

<h3>Summary</h3>

This change adds capability tokens to Telnyx event callbacks to prevent unauthenticated run enumeration. It must not merge until it preserves valid callbacks for calls that were already in flight when token enforcement is enabled.

<sub>Reviews (2) · Last reviewed commit: ["fix: return uniform 404 for invalid teln..."](https://github.com/dograh-hq/dograh/commit/307bff1ff5c367db6a23ef2299ac0aebdd51eaa3)</sub>


a692570:
Addressed the Greptile timing finding in a follow-up commit (307bff1f):

**What we changed.** When `TELEPHONY_WS_TOKEN_SECRET` is configured, the events webhook URL minted by the Telnyx provider now carries an HMAC capability token (`ws_auth.mint_events_token`, bound to the run id, same secret and style as the existing media-WS token). `handle_telnyx_events` verifies it with a constant-time compare before any database lookup: missing or invalid token returns the same uniform 404 immediately, with zero pre-auth DB work. With no secret configured, the flow keeps the lookup-then-verify order documented in this PR.

**Why this design.** The reviewer's suggested fixes were (a) verify authenticity before run-dependent lookups, or (b) equalize work on the missing-run path. (b) is not implementable cleanly, since the auth key is org-scoped and the org is only derivable from the run. (a) is now real: the token is request-level, minted at call time by the same code path that built the webhook URL, and verification is pure CPU. An unsigned probe does path-param parse plus one HMAC compare regardless of run existence, so both the status-code and timing channels close.

**Transport note.** The repo's media-WS docs warn that carriers strip query strings on stream URLs, so a query-param token looked unsafe at first. We verified live against the Telnyx API (2026-09-16): Telnyx POSTs webhook callbacks to the URL verbatim, query string intact. The media-WS restriction applies to `stream_url`, not `webhook_url`, so `?token=` is safe here.

**Tests.** New `api/tests/telephony/telnyx/test_events_capability_token.py` (7 tests): provider mints the URL with the token when the secret is set and without it when unset; the route rejects missing/invalid tokens before any DB lookup (asserted via mocks), accepts valid tokens, and degrades to the legacy flow with no secret. `ws_auth` gained events-token round-trip/rejection tests. Full api suite green on the branch (3252 passed, 1 skipped).

Known limitation from the original PR description still stands: #786's 400 for provider-resolution failures remains distinguishable after the gate; absorbing that is follow-up work.

a6kme:
Pushed `e947bd5f2` addressing the blocking rollout issue plus several review findings.

**The blocker: this would have leaked campaign call slots on deploy.**

`TELEPHONY_WS_TOKEN_SECRET` is already set in production, so `if ws_auth.token_configured():` meant the gate went hard-enforcing the moment this deployed. Every call in flight across the rolling deploy registered its Telnyx `webhook_url` *before* the token existed, so its `call.hangup` would 404 at the gate — never reaching `_process_status_update`, and therefore never calling `campaign_call_dispatcher.release_call_slot()`. Those campaign concurrency slots leak permanently. The run also stays non-`COMPLETED` and the circuit breaker never records.

Fixed by following the two-phase pattern this module already established for the media socket: new `TELEPHONY_EVENTS_TOKEN_ENFORCE`, defaulting off. Until it's flipped, a missing/invalid token is logged as `UNVERIFIED Telnyx events request` and falls through to signature verification — which is the real authentication either way. It needs its own flag rather than reusing `TELEPHONY_WS_TOKEN_ENFORCE` because the events URL is lodged with the carrier at dial time and outlives a deploy, unlike a media socket.

**The tests could not run in the production configuration.**

`test_events_route_uniform_404.py` (and the pre-existing `_request` helper in `test_provider.py`) built ASGI scopes with no `query_string` key. Starlette indexes `scope["query_string"]` directly, so with the secret set all three uniform-404 tests died with `KeyError: 'query_string'` — they only ever passed because `api/.env.test` leaves the secret unset. Worse, with enforcement on, every request short-circuits at the gate and the 401→404 signature-path change this PR is named after would have been unreachable in production. Added `query_string` to the scopes and a `test_uniform_404s_hold_in_the_production_configuration` that pins the behaviour with a secret set.

**The capability-token tests were vacuous.** `_post_with_token` stubbed `get_workflow_run_by_id` to return `None`, which itself yields `404 "Workflow run not found"` — identical to the reject path. `test_valid_token_passes_the_gate` would have stayed green if the gate rejected every valid token. It now returns a real run and asserts `assert_not_awaited()` / `assert_awaited_once()` on the lookup, which is what actually distinguishes the two paths.

**Also in this push**

- The run-existence channel was still open for one case: a run whose workflow row is gone returned `404 "Workflow not found"`, byte-distinguishable from `404 "Workflow run not found"`. Now uniform.
- `json.loads(raw_body)` was unguarded — a malformed body was an uncaught 500 with a stack trace on a path that is unauthenticated until the signature check. Now a 400.
- `log_configuration_status()` reported only the media socket, so an operator would read "enforcement off — invalid tokens are logged and still allowed" while event webhooks were being rejected. It now reports both surfaces.
- Documented `TELEPHONY_EVENTS_TOKEN_ENFORCE` in `environment-variables.mdx` and `.env.example`, and noted in `webhooks.mdx` why this token rides in a query string when the media-socket one deliberately does not — that doc previously stated carriers can't be trusted with query strings, with no mention of the Telnyx webhook exception.
- `./scripts/format.sh` (this was the red `drift-check`).

**Not addressed, worth a follow-up:** the token is an HMAC over the run id with no expiry, and it lands in the nginx/uvicorn access log on every event — 5-10 log lines per call, shipped to Axiom, where `redact_token()` does not run. Anyone with log read access can replay the gate for any run they can see. `build_media_ws_url` avoids this by using a path segment. Given the gate is defence-in-depth behind signature verification, I left it, but a timestamped token or a path segment would close it.

Also unrelated but adjacent: `/telnyx/transfer-result/{transfer_id}` in this same file still has no signature verification and no token at all.

